PR #11, opened 2013-04-03
Cinder volumes group name

dmi-try:


Miroslav-Anashkin:
I do not think it is good idea to set default cinder user/password as class input parameter default values.

PR #17, opened 2013-04-04
Mirrorlist changes, version info implement 

ashaposhnikov:
All mirrorlist definitions removed except mirantis and puppetlabs repos.
Version added to the iso name, version.yaml at root of iso and to the /etc/fuel/version.yaml


xenolog:
I can't understand what and why makes. :(

PR #35, opened 2013-04-11
allow repos customization

aglarendil:


deployment/puppet/cobbler/examples/server_site.pp
```diff
@@ -16,7 +16,16 @@
 $pxetimeout          = '0'
 $dhcp_interface      = 'eth1'
 $mirror_type         = 'default'
-
+$fast_mirror         = 'http://172.18.67.168'
```
aglarendil, line 19: fast_mirror is just a prefix here. we then concatenate with specific path

aglarendil, line 19: These lines differ only by distro-specific path. I cannot specify full path

PR #61, opened 2013-04-26
fix cinder string or array check

mattymo:
Fixes this error: Error: Could not retrieve catalog from remote server: Error 400 on SERVER: member(): Requires array to work with at /etc/puppet/modules/openstack/manifests/controller.pp:191 on node fuel-controller-01.your-domain-name.com\x1b[0m\n', '\x1b[1;31mError: Could not retrieve catalog; skipping run


mattymo:
Victor, how can we determine the role of a node (storage, controller, compute) from site.pp without writing extra code to parse hostname? I would need many more lines of puppet code to move this logic all to site.pp.

PR #76, opened 2013-05-07
Fix missed nv_physical_volume for simple:compute class

bogdando:
Fixes https://mirantis.jira.com/browse/FUEL-597


bogdando:
if cinder = ['computes' / 'all'] selected for simple manifest it wouldn't have provided cinder-volumes vg @ computes unless nv_physical_volume included for its compute class

mattymo:
Looks good to me

PR #131, opened 2013-05-18
Rsyslog: saveable Disk-Assisted Memory Queues, async writes, batching, rate-limiting

bogdando:
- Async writes to syslog
- Disk-Assisted Memory Queues
- Number of queue workers 1->2
- Save queues on shutdown/exit to ensure no messages was lost
- Rate limiting (no more than 1,000 messages within a second)
- Queue messages batching 16->128 (might improve performance)
- Do not control escape characters on receive (http://www.rsyslog.com/doc/rsconf1_escapecontrolcharactersonreceive.html)

TODO imporve fixed work dir names for DA queues /var/lib/rsyslog

Links:
http://www.rsyslog.com/doc/queues.html
http://www.rsyslog.com/doc/rsyslog_conf_global.html


deployment/puppet/rsyslog/templates/rsyslog.conf.erb
```diff
@@ -23,6 +23,25 @@ $PrivDropToUser <%= scope.lookupvar('rsyslog::params::run_user') %>
 $PrivDropToGroup <%= scope.lookupvar('rsyslog::params::run_group') %>
 
 #
+# Disk-Assisted Memory Queues, async writes, no escape chars
+#
+$OMFileASyncWriting on
+$EscapeControlCharactersOnReceive off
```
xenolog: Do you shure???  You want see color messages in logs???

bogdando: What colour messages? I cannot find any relation to this here http://www.rsyslog.com/doc/rsconf1_escapecontrolcharactersonreceive.html or here http://www.rsyslog.com/doc/rsyslog_conf_global.html

xenolog: EscapeControlCharactersOnReceive
This directive instructs rsyslogd to replace _control_characters_ during reception of the message. The intent is to provide a way to stop non-printable messages from entering the syslog system as whole.

PR #132, opened 2013-05-18
CI Add DEBUG for pupet

bogdando:


bogdando:
Can you clarify please?
Running "puppet agent -tvd --evaltrace" instead of "puppet agent --test" was quite useful in grizzly research, it provides more information for troubleshooting, including commands issued & time evaluation for each action taken. 
For example, it might be useful if puppet agents applying taking a long time

PR #137, opened 2013-05-21
Ci libvirt type

bogdando:
Fixes https://mirantis.jira.com/browse/FUEL-645


bogdando:
Yes, it works indeed. Cannot reproduce the case with libvirt_type=kvm causing "unknown OS 
type hvm" error.

PR #147, opened 2013-05-22
Docs: Initial advanced config section for 2.2

nickchase:
Includes stub for Quantum HA and raw material for NIC bonding.  Custom plugins is ready for review.


plechenko:
Please change all domain names to .localdomain

PR #175, opened 2013-05-24
FIX: give more information while stupid exception

xenolog:


fuel_test/helpers.py
```diff
@@ -40,10 +40,14 @@ def extract_virtual_ips(ipaout):
 
 
 def write_config(remote, path, text):
-    config = remote.open(path, 'w')
-    config.write(text)
-    logging.info('Write config %s' % text)
-    config.close()
+    try:
+        #todo: move to down layer (paramiko)
```
xenolog, line 44: By default we have following information:
    IOError: [Errno 2] No such file
No remote IP address, no filename given.

I re-raise (at border between our code and paramiko library) the same exception  with more powerful information message for deployment engineer  convenience.

xenolog, line 44: In deployment process we want know which file can't be written, and on which node.
This information more interesting for debugging,  than line number in 3d-party libruary.

PR #199, opened 2013-05-28
Add pungi iso format

mattymo:
This method allows make iso ISO_METHOD=pungi which will generate a full iso via pungi, including all packages required for Fuel deployment. Old style "make iso" still works as before.


iso/ks-full.cfg
```diff
@@ -0,0 +1,236 @@
...
+cronie
+puppet-2.7.19
+puppet-server-2.7.19
+puppetdb
+python-argparse
+mcollective
+mcollective-client
+rubygem-astute
+ruby-devel
+gcc
```
xdel: we may safely threw out both gcc and make, they are not needed at all

PR #203, opened 2013-05-29
Fixes for grizzly support

aglarendil:


bogdando:
Please hold a merge, I need to investigate it, looks like this is a step back to old cinder issues were discovered while research...

fuel_test/manifest.py
```diff
@@ -237,7 +237,7 @@ def generate_openstack_manifest(self, template,
                                     quantum=True,
                                     loopback=True,
                                     cinder=True,
-                                            cinder_nodes=None,
+                                            cinder_nodes=[],
```
aglarendil, line 240: if it is set to None - it converts to undef in site.pp, which breaks installation. anyway it's value should be list, not NoneType.

PR #205, opened 2013-05-29
dhcp checksum fix

xdel:


bogdando:
Please provide this fix for all nodes, not master node only. It will fix dhcp issues for instances as well. 
And send it to grizzly branch also.

xdel:
I need to test it on already deployed env - it should be easier to switch network driver rather than putting very specific workaround everywhere.

PR #210, opened 2013-05-30
Two fixes - puppet and iptables from master

xdel:


xdel:
Invalid cause of the problem. Not for merge.

Should be fixed by adding rule manually at the host.

PR #212, opened 2013-05-30
Grizzly fixes

aglarendil:
Revert to UUID in keystone due to OSCI-363 bug.
Save compatibility with nova-volume for Folsom.
Misc tests fixes.


bogdando:
Can you please provide an example of issues with PKI mode running grizzly? 
osci-363 is folsom related only.

PR #228, opened 2013-06-04
Grizzly  Quantum  vlan mode

xenolog:
tested on libvirt VMs


xenolog:
tested on bare metal  in 'simple' variant.

PR #229, opened 2013-06-04
Added explicit new Galera cluster start procedure

Miroslav-Anashkin:
Changed default template for mysql galera. 
Removed empty gcomm. 
Added explicit new Galera cluster start procedure, based on UUID value in /var/lib/mysql/grastate.dat. 
Removed SElinux disable code - it should not exist in galera manifest only but it should work on every node.


plechenko:
+1

PR #234, opened 2013-06-05
FUEL-688

demofly:


aglarendil:
We do already have this change in our code

PR #238, opened 2013-06-05
Corrected starting condition for selinux disable sequence 

Miroslav-Anashkin:
Actually all facts are strings, so all conditional operators should treat them as strings.


plechenko:
Is it out of date?

PR #239, opened 2013-06-06
set glance endpoint everywhere; hardcode auth mech to keystone

xdel:


deployment/puppet/openstack/manifests/cinder.pp
```diff
@@ -9,6 +9,7 @@
   $physical_volume = undef,
   $manage_volumes  = false,
   $enabled         = true,
+  $glance_connection = ['127.0.0.1'],
```
bogdando: glance_connection or glance_api_servers maybe?

PR #248, opened 2013-06-10
more appropriate workaround

xdel:
Works better than looks. Needs to be tested with swift.


bogdando:
This renders wrong glance_api_servers value for controller node:
[root@fuel-controller-01 ~]# grep glance_ /etc/cinder/cinder.conf 
glance_api_servers=glance_api_servers
[root@fuel-controller-01 ~]# ssh fuel-compute-01 grep glance_ /etc/cinder/cinder.conf 
glance_api_servers=10.108.12.5:9292
[root@fuel-controller-01 ~]# ssh fuel-compute-02 grep glance_ /etc/cinder/cinder.conf 
glance_api_servers=10.108.12.5:9292

PR #249, opened 2013-06-11
libvirt proposal

xdel:
Please test everywhere.

### DIFF
--- a/deployment/puppet/nova/manifests/compute/libvirt.pp
+++ b/deployment/puppet/nova/manifests/compute/libvirt.pp
@@ -40,6 +40,14 @@ class nova::compute::libvirt (
 
     Service['avahi-daemon'] -> Service['libvirt']
 
+    service { 'libvirt-guests':
+      name       => 'libvirt-guests',
+      enable     => false,
+      ensure     => true,
+      hasstatus  => false,
+      hasrestart => false,
+    }
+
   }
 
   Service['libvirt'] -> Service['nova-compute']


deployment/puppet/nova/manifests/compute/libvirt.pp
```diff
@@ -40,6 +40,14 @@
 
     Service['avahi-daemon'] -> Service['libvirt']
 
+    Exec['libvirt-guests-off'] -> Service <| title == 'libvirt' |>
+
+    exec {'libvirt-guests-off':
```
xenolog: -1

why we do not use the Service resource?

deployment/puppet/nova/manifests/compute/libvirt.pp
```diff
@@ -49,6 +57,13 @@
       ensure => present,
       before => Package[$::nova::params::compute_package_name],
     }
+    service { 'libvirt-guests':
```
aglarendil: Ubuntu part is not needed. Remove it

deployment/puppet/nova/manifests/compute/libvirt.pp
```diff
@@ -40,6 +40,14 @@
 
     Service['avahi-daemon'] -> Service['libvirt']
 
+    service { 'libvirt-guests':
+      name       => 'libvirt-guests',
+      enable     => false,
+      ensure     => true,
```
aglarendil, line 46: why this "ensure" is set to true ??

bogdando, line 46: 'stopped' maybe?

xdel, line 46: I had explained this before - script is completely stateless, so removing symlinks from every runlevel will do the same.

PR #259, opened 2013-06-14
Add Qpid module to cluster HA.

dkalashnik:


deployment/puppet/openstack/manifests/controller_ha.pp
```diff
@@ -321,6 +330,12 @@
       rabbit_port             => '5673',
       rabbit_node_ip_address  => $rabbit_node_ip_address,
       rabbit_ha_virtual_ip    => $internal_virtual_ip,
+      qpid_password           => $rabbit_hash[password],
+      qpid_user               => $rabbit_user,
+      qpid_cluster            => true,
+      qpid_nodes              => $controller_hostnames,
+      qpid_port               => '5673',
```
Miroslav-Anashkin: We use port 5673 for HAProxy virtual IP on HAProxy RabbitMQ frontend. RabbitMQ backends (nodes) are always use 5672
So, what was the reason of setting port to 5673 if no QPid service configured behind HAProxy exist?

dkalashnik: Fixed.

PR #270, opened 2013-06-18
Fuel bugs for 3.0.1

aglarendil:


deployment/puppet/cinder/manifests/params.pp
```diff
@@ -18,10 +18,11 @@
       $tgt_package_name  = 'tgt'
       $tgt_service_name  = 'tgt'
       $python_path       = 'python2.7/dist-packages'
-
+      $qemuimg_package_name = 'qemu-utils'
     }
 
     'RedHat': {
+      $qemuimg_package_name = 'qemu-img'
```
xdel, line 25: WHAT

qemu-kvm-tools

deployment/puppet/rsyslog/templates/client.conf.erb
```diff
@@ -11,6 +11,7 @@ $ActionQueueMaxDiskSpace 1g     # 1gb space limit (use as much as possible)
 $ActionQueueSaveOnShutdown on   # save messages to disk on shutdown
 $ActionQueueType LinkedList     # run asynchronously
 $ActionResumeRetryCount -1      # infinety retries if host is down
+$SystemLogRateLimitInterval 0   # disable rate limits for rsyslog
```
bogdando, line 14: Probably it better should be, for example 
$SystemLogRateLimitInterval 2
$SystemLogRateLimitBurst 100
which means not more than 100 messages in 2 seconds

PR #301, opened 2013-07-05
Add colocation for public-vip to p_mysql

ilyinon:
It adds colocation public-vip to p_mysql

### DIFF
--- a/deployment/puppet/mysql/manifests/server.pp
+++ b/deployment/puppet/mysql/manifests/server.pp
@@ -151,6 +151,14 @@ class mysql::server (
       require    => [Cs_resource['p_mysql'], Cs_commit['mysql']],
     } 
 
+    #Tie public-vip to p_mysql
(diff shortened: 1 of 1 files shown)


deployment/puppet/mysql/manifests/server.pp
```diff
@@ -151,6 +151,14 @@
       require    => [Cs_resource['p_mysql'], Cs_commit['mysql']],
     } 
 
+    #Tie public-vip to p_mysql
+    cs_colocation { 'mysql_to_public-vip':
```
xenolog, line 155: Why use PUBLIC VIP instead PRAVATE (management) ?

mattymo, line 155: There is already colocation to private.

xenolog, line 155: I think Mysql do not listen on public interfaces. it's insecure.

PR #309, opened 2013-07-08
Undo fix for FUEL-546

bogdando:
net04_ext should be created as non shared, as it was before FUEL-546.

### DIFF
--- a/deployment/puppet/quantum/manifests/agents/l3.pp
+++ b/deployment/puppet/quantum/manifests/agents/l3.pp
@@ -145,7 +145,7 @@ class quantum::agents::l3 (
         subnet_gw       => $external_gateway, # undef,
         alloc_pool      => $external_alloc_pool, # undef,
         enable_dhcp     => 'False', # 'True',
-        shared          => 'True',
+        shared          => 'False',
       }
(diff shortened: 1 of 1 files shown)


aglarendil:
And why should we make this network unshared ?

bogdando:
It was before for some reasons.
git diff stable/master..2.2-folsom  deployment/puppet/quantum/manifests/agents/l3.pp
--- a/deployment/puppet/quantum/manifests/agents/l3.pp
+++ b/deployment/puppet/quantum/manifests/agents/l3.pp
@@ -145,7 +145,6 @@ class quantum::agents::l3 (
         subnet_gw       => $external_gateway, # undef,
         alloc_pool      => $external_alloc_pool, # undef,
         enable_dhcp     => 'False', # 'True',
-        shared          => 'True',
       }
       Quantum_l3_agent_config <| |> -> Quantum::Network::Setup['net04_ext']

mrasskazov:
When net04_ext has Shared=True, instance receives ip address from net04_ext by default, i.e. floating ip. It is not a standard case. We must provide network manually in this case: nova boot ... --image ... --flavor ... --nic net-id=<NET_ID>. The Tempest is not able to set net-id when it creates instances. As a result floating ips are over in net04_ext and new instances become in ERROR state.

PR #318, opened 2013-07-10
Patch 2

mattymo:


mattymo:
Not functioning and missing qpid definitions completely in controller.pp and controller_ha.pp

PR #335, opened 2013-07-18
Fuel 3.0.1 rhosdev

dkalashnik:
Add after-merge fixes


mattymo:
Thanks for fixing typo and setting unicast corosync

PR #341, opened 2013-07-18
Override yum provider core plugin for puppet

bogdando:
Ensures yum provider would have expired metadata cache
and redo last, if operation had failed
scope: puppet 2.7.19

Implements: #FUEL-632

### DIFF
--- a/deployment/puppet/puppet/lib/puppet/provider/package/yum.rb
+++ b/deployment/puppet/puppet/lib/puppet/provider/package/yum.rb
@@ -1,0 +1,121 @@
+require 'puppet/util/package'
+
+Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
+  desc "Support via `yum`."
+
+  has_feature :versionable
+
+  commands :yum => "yum", :rpm => "rpm", :python => "python"
+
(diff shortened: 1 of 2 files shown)


deployment/puppet/puppet/lib/puppet/provider/package/yum.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    begin
+      limiter += 1
+      yum "-d", "0", "-e", "0", "-y", operation, wanted
+    rescue Puppet::ExecutionFailure => error
+      debug("Got error while operation:#{operation}, wanted:#{wanted}:\n#{error}")
+      if limiter <= tries
+        debug("Cleaning yum cache and retrying operation, try #{limiter} of #{tries}...")
+        yum "clean", "expire-cache"
+        #yum "history", "redo", "last"
```
bogdando, line 83: That is just the memo, no need to execute because of 'retry' below...

mattymo:
This will help if we can't just patch puppet RPM. Either way, it will help us with short-lived repo issues.

bogdando:
not applicable since fuel 3.1 consolidated iso

PR #353, opened 2013-07-22
sync site.pp between configurations

xenolog:
- FIX virtual_ip for non-quantum modes

### DIFF
--- a/deployment/puppet/openstack/examples/site_openstack_ha_compact.pp
+++ b/deployment/puppet/openstack/examples/site_openstack_ha_compact.pp
@@ -26,17 +26,6 @@ $internal_virtual_ip = '10.0.0.253'
 # interface resides
 $public_virtual_ip   = '10.0.204.253'
 
-$vips = { # Do not convert to ARRAY, It's can't work in 2.7
-  public_old => {
-    nic    => $public_br,
-    ip     => $public_virtual_ip,
-  },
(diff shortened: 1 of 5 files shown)


xenolog:
do not merge without QA.

deployment/puppet/openstack/examples/site_openstack_ha_compact.pp
```diff
@@ -496,7 +498,7 @@
 # In case of non debug and non verbose - WARNING, default level would have set.
 # Note: if syslog on, this default level may be configured (for syslog) with syslog_log_level option.
 $verbose = true
-$debug = false
+$debug = true
```
bogdando, line 501: I believe, we should consider standard mode as verbose only. Debug would generate too much information for production deployments...

deployment/puppet/openstack/examples/site_openstack_simple.pp
```diff
@@ -407,7 +407,7 @@
 # In case of non debug and non verbose - WARNING, default level would have set.
 # Note: if syslog on, this default level may be configured (for syslog) with syslog_log_level option.
 $verbose = true
-$debug = false
+$debug = true
```
bogdando, line 410: I believe, we should consider standard mode as verbose only. Debug would generate too much information for production deployments...

PR #371, opened 2013-07-24
return GRE isolation by default

xdel:


bogdando:
Please, do not forget about compact_for_docs example as well

PR #373, opened 2013-07-25
powerfull debug logging for Quantum L3 agent

xenolog:


bogdando:
This one is related to FUEL-835. We should also investigate same options for non-HA quantum case and for the rest of openstack services as well

bogdando:
This patch won't work correct unless /etc/init.d/quantum-\* would be fixed to redirect daemon output to stderr.
With default init script for quantum-server, it would not provide any logs to /var/log/quantum/server.log

PR #387, opened 2013-07-26
Make yum not so fat

xdel:


deployment/puppet/cobbler/templates/kickstart/rhel.ks.erb
```diff
@@ -117,6 +117,7 @@ $SNIPPET('mcollective_install_if_enabled')
 yum versionlock puppet
 yum versionlock kernel
 yum versionlock iproute2
+sed -i 's/enabled=1/enabled=0/g' /etc/yum/pluginconf.d/fastestmirror.conf
```
mattymo: yum-config-manager --disableplugin=fastestmirror --save  &>/dev/null

PR #388, opened 2013-07-26
Fixes for migration L3 agents in HA mode

xenolog:
works only with patched L3 agent.

### DIFF
--- a/deployment/puppet/quantum/files/ocf/quantum-agent-l3
+++ b/deployment/puppet/quantum/files/ocf/quantum-agent-l3
@@ -356,7 +356,7 @@ quantum_l3_agent_start() {
         sleep 1
     done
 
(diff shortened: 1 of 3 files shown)


deployment/puppet/quantum/files/ocf/quantum-agent-l3
```diff
@@ -356,7 +356,7 @@ quantum_l3_agent_start() {
         sleep 1
     done
 
-    sleep 7 ; q-agent-cleanup.py --agent=l3 --reschedule --remove-dead 2>&1 >> /var/log/quantum/rescheduling.log
+    sleep 13 ; q-agent-cleanup.py --agent=l3 --reschedule --remove-dead 2>&1 >> /var/log/quantum/rescheduling.log
```
xarses, line 359: Why are we sleeping / increasing the sleep here? Can't we loop until the service is up usable instead of static sleep staments?

PR #393, opened 2013-07-29
Multiboot fix2

kozhukalov:


aglarendil:
Pls, check for any regressions possible (e.g. that the simplest ks_spaces with only one disk as PV will work)

PR #396, opened 2013-07-29
Fix mysql tests

dmitryilyin:


deployment/puppet/mysql/spec/integration/default/backup_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+end
+
+describe file('/usr/local/sbin/mysqlbackup.sh') do
+  it { should be_file }
+  it { should be_mode 700 }
+  it { should be_owned_by 'root' }
+  it { should be_grouped_into 'root' }
+end
+
+describe file('/tmp/backups') do
```
xarses, line 15: Why are we writing backups to /tmp? /tmp is erased automaticly, often during reboot. http://askubuntu.com/questions/20783/how-is-the-tmp-directory-cleaned-up http://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard

xdel, line 15: Centos does not following even those, so /tmp are preserved across reboots, lol. Of course there should be more reliable way as /var/tmp.

xarses, line 15: for RHEL/CentOS this will occur if tmpwatch is installed (and files are unaccessed for a specific age). regardless, this is not a good place for data that we want to persist.

dmitryilyin, line 15: I really don't know why /tmp/backups was chosen as default backups location.
Looks like author of this module was expecting a user to override this path with his own.
Anyway it doesn't matter because we don't use mysql::backup.

PR #400, opened 2013-07-29
New floating IP range support

aglarendil:


deployment/puppet/nova/manifests/manage/floating.pp
```diff
@@ -1,3 +1,4 @@
+# This class nova::manage::floating has been deprecated in favor of nova::manage::floating_range
 define nova::manage::floating (
```
mihgen, line 2: Why not to remove it if it's deprecated?

PR #404, opened 2013-07-29
Remove double declaration

bogdando:
would ignore puppet_master_log setting and create rsyslog rule for puppet-master log
if this option should not be ignored, do not merge it (rsyslog::server declaration should be removed then)


deployment/puppet/puppetmaster/manifests/master.pp
```diff
@@ -23,14 +23,15 @@
     notify => Service["puppetmaster"],
   }
 
-  if $puppet_master_log == "syslog" {
```
bogdando, line 26: would ignore puppet_master_log setting and create rsyslog rule for puppet-master log

PR #413, opened 2013-07-30
disable haproxy for rabbitmq epmd service

aglarendil:


Miroslav-Anashkin:
+1
EPMD does not need HAProxy

PR #417, opened 2013-07-30
Syslog message format fuelweb compatibility fixes

bogdando:
Syslog message format fuelweb compatibility fixes
- Add semicolon after severity
- Use date-rfc3339 for rsyslog server


adanin:
+1

PR #419, opened 2013-07-30
[rpmcache] Restore mysql packages.

adanin:


mattymo:
-1 we do not want MySQL 5.5 for RHEL. It is not supported.

PR #423, opened 2013-07-30
fix mysql service name for centos/rhel

mattymo:
Need to set mysql to mysqld. Also fix corosync refs to use p_${service_name}


dmi-try:
+1

xdel:
+1

PR #424, opened 2013-07-30
Fix cobbler template

dmi-try:


mattymo:
Fixes cheetah's penultimate hatred for dollar signs it doesn't own.

PR #430, opened 2013-07-31
Add missed puppet-agent log config for rsyslog server

bogdando:
- Restore rsyslog config for puppet-agent logs at master node (rsyslog server)


mihgen:
verified by manually placing the file.

PR #436, opened 2013-07-31
Add openstack logging client to replace rsyslog::client

bogdando:
- Configure openstack logging client role for cluster nodes

need for testing new logging system with fuelweb deploy
would conflict with Vladimir Kuklin's solution for sure... 


aglarendil:
I suggest not to merge it, because we are going to have the same code in site.pp after merge with consolidated code.

aglarendil:
-1

PR #443, opened 2013-08-01
define floating ips range

Zipfer:


aglarendil:
I need Dima Pyzhov's acceptance

PR #449, opened 2013-08-02
Powerfull debug mode for openstack syslogging

bogdando:
Implements option 2 for FUEL-835 suggested solutions
Workarounds for FUEL-842, FUEL-843
- StdErr must not be used unless all quantum+agents init & ocf scripts would redirect its stderr to logfiles (&>>/var/log/foo/bar.log)
- Thus, undo powerfull debug via stderr - #373
- Remove use_stderr, use logdir options for stdout instead
- Dirty hack for quantum init script to remove all its --log-config from daemon, cuz it breaks logging
- Use rsyslog::imfile monitoring features for local openstack logfiles for syslog + debug case
- Fix imfile templates for quantum debug to use normalized syslogtags
- Put verbose/debug section on top of site.pp to fix top scope definitions issues for logging class
- Get rid of global-scope referenced variables for swift:: and rsyslog::client
- High precision timestamps (date-rfc3339) as default
- Weekly rotate & keep rotated logs for 4 weeks at master node as default


bogdando:
not applicable to develop anymore

PR #455, opened 2013-08-02
Fix logclient 777vvkbased

bogdando:
Port fix-835, wa-842, wa-843 from fuel-lib for powerfull debug mode


bogdando:
Testing this (fix_logclient_777vvkbased) @ fuelweb, please hold for a while

PR #463, opened 2013-08-05
fix sat_hostname ksmeta for non-satellite case

mattymo:
Fixes this issue:
err:  Failed to apply catalog: Parameter ksmeta failed: Invalid value "redhat_register_user=USERNAME redhat_register_password=PASSWORD redhat_management_type=cert redhat_management_server= ". Valid values match /^((\S+=\S+) +)_(\S+=\S+)_$/. at /etc/puppet/modules/rpmcache/manifests/init.pp:112


adanin:
+1

PR #469, opened 2013-08-05
Add cleanup step to rpmcache and enable nailgun repo

mattymo:


xdel:
+1
looks good

deployment/puppet/rpmcache/templates/build_rpm_cache.erb
```diff
@@ -1,4 +1,32 @@
-#!/bin/bash -x
+#!/bin/bash -xe
+
+function revert_back_to_centos() {
+  # Removes redhat-release-server rpm and replaces it with centos-release
+  echo "Removing Red Hat temporary configuration."
+  if rpm -q redhat-release-server &>/dev/null; then
+    rpm -Uvh --force /root/centos-release/centos-release*.rpm
```
mihgen, line 7: Who is creating this RPM there?

deployment/puppet/rpmcache/templates/build_rpm_cache.erb
```diff
@@ -63,12 +89,18 @@ fi
...
+  sed -i 's/^enabled.*/enabled=0/' "/etc/yum/pluginconf.d/${plugin}.conf" 2>/dev/null || :
+done
+
 #Run registration
 
 rhnreg_ks "--serverUrl=http://<%= sat_hostname %>/XMLRPC" --profilename=fuelweb "--activationkey=<%= activation_key %>" --sslCACert=/etc/sysconfig/rhn/RHNS-CA-CERT --nopackages --nohardware --novirtinfo --norhnsd --force
 
 if [ $? -ne 0 ]; then
   echo "Registration failed. Aborting."
+  revert_back_to_centos
```
adanin: As I understand it will be run when you call 'exit 1' because of 'trap revert_back_to_centos EXIT'. You don't need to call it explicitly.

deployment/puppet/rpmcache/templates/build_rpm_cache.erb
mattymo, line 7: This is created and installed if you do RHN Satellite and add redhat-release-server at line 75. We need to switch back to centos-release rpm at the end.

deployment/puppet/rpmcache/templates/build_rpm_cache.erb
mattymo: Okay. I will remove them all :(

PR #474, opened 2013-08-05
Disable puppet reports causing agent hangs

dmitryilyin:


xarses:
Does puppetdb need these for stored configs?
+1 if it dosn't, this process takes 10-20% of the agent runtime.

mihgen:
+1

PR #476, opened 2013-08-06
Prepare repos for outside usage

xdel:


bogdando:
:-1: 
I believe that should be ON/OFF by enable_test_repo option instead.
f.e. like this:https://github.com/Mirantis/fuel/pull/450

PR #477, opened 2013-08-06
Add /var/log/syslog to time based rotation as well

bogdando:
/var/log/syslog have been initially added to size based rotaion plan (20-fuel.conf) only.
Should have been added to time based rotation plan (10-fuel.conf) as well.


bogdando:
redirect to develop

PR #480, opened 2013-08-06
Fix logclient

bogdando:
Implements option 2 for 835 suggested solutions
WA for FUEL-842, FUEL-843
- StdErr must not be used unless all quantum+agents init & ocf scripts would redirect its stderr to logfiles (&>>/var/log/foo/bar.log)
- Thus, undo powerfull debug via stderr - #373
- Remove use_stderr, use logdir options for stdout instead
- Dirty hack for quantum init script to remove all its --log-config from daemon, cuz it breaks logging
- Use rsyslog::imfile monitoring features for local openstack logfiles for syslog + debug case
- Fix imfile templates for quantum debug to use normalized syslogtags
- Get rid of top scoped global vars references (for logging features)
- Add /var/log/{syslog,corosync} to rotation plans
- Fix verbose/debug logic for cluster_*, examples, and site.pp
- High timestamps date-rfc3339 as default
- Weekly rotate & keep rotated logs for 4 weeks at master node as default (was daily, keep 7)

### DIFF
--- a/deployment/puppet/cinder/manifests/base.pp
+++ b/deployment/puppet/cinder/manifests/base.pp
@@ -4,8 +4,7 @@
 # $osapi_volume_extension = cinder.api.openstack.volume.contrib.standard_extensions
 # $root_helper = sudo /usr/local/bin/cinder-rootwrap /etc/cinder/rootwrap.conf
 # $use_syslog = Rather or not service should log to syslog. Optional.
-# $syslog_log_facility = Facility for syslog, if used. Optional. Note: duplicating conf option
-#       wouldn't have been used, but more powerfull rsyslog features managed via conf template instead
+# $syslog_log_facility = Facility for syslog, if used. Optional.
 # $syslog_log_level = logging level for non verbose and non debug mode. Optional.
 
 class cinder::base (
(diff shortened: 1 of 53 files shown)


deployment/puppet/cinder/manifests/base.pp
```diff
@@ -72,16 +77,30 @@
...
+     value => '%(asctime)s %(levelname)s %(name)s [-] %(instance)s %(message)s';
+  }
+  # might be used for stdout logging instead, if configured
+  file { "cinder-logging.conf":
+    content => template('cinder/logging.conf-nosyslog.erb'),
+    path => "/etc/cinder/logging.conf",
+    require => File[$::cinder::params::cinder_conf],
+  }
 }
+  # We must notify services to apply new logging rules
```
aglarendil, line 97: use variables $::cinder::params::$api_service|$scheduler_service|$volume_service for compatibility with Ubuntu

deployment/puppet/quantum/manifests/init.pp
```diff
@@ -186,23 +199,30 @@
...
       path  => "/etc/quantum/logging.conf",
       owner => "root",
-      group => "root",
-      mode  => 644,
+      group => "quantum",
+      mode  => 640,
     }
   }
-
+  # We must setup logging before start services under pacemaker
```
aglarendil, line 218: use also corresponding $::quantum::params::<service_name>_service or similar for compatibility with Ubuntu

PR #503, opened 2013-08-23
Add CEPH manifests

xarses:
Adds manifests to deploy Ceph parts.


xarses:
@xdel the ceph-deploy script makes extensive use of sudo, even though its running as root. Also the ceph-deploy scripts require ssh-keys ceph/files/openstack\* that need to be on every controller, storage and compute.

xarses:
@xdel yum.pp isn't called currently because the packages are in the repo and is currently not being called.

deployment/puppet/ceph/manifests/deploy.pp
```diff
@@ -0,0 +1,89 @@
...
+    exec { 'ceph-deploy-s4':
+      command => "ceph-deploy mds create ${mds}",
+      require => Class['c_osd'],
+      logoutput => true,
+    }
+  }
+  class c_pools {
+    define int {
+      exec { "Creating pool ${name}":
+        command => "ceph osd pool create ${name} 128",
```
xdel: pg_num and pgp_num should be multiplied by number of computes

deployment/puppet/ceph/manifests/cinder.pp
```diff
@@ -0,0 +1,52 @@
...
+      hasstatus  => true,
+      hasrestart => true,
+    }
+    exec { 'Create keys for pool volumes':
+      command => 'ceph auth get-or-create client.volumes > /etc/ceph/ceph.client.volumes.keyring',
+      before  => File['/etc/ceph/ceph.client.volumes.keyring'],
+      require => [Package['ceph'], Exec['Copy configs']],
+      returns => 0,
+    }
+    file { '/etc/ceph/ceph.client.volumes.keyring':
```
xdel, line 49: please put safer permissions on every keyring not created by ceph-deploy

deployment/puppet/ceph/examples/site.pp
```diff
@@ -0,0 +1,83 @@
...
+      chain => 'INPUT',
+      dport => '6800-7100',
+      proto => 'tcp',
+      action  => accept,
+    }
+  }
+  if $fqdn == $mon_nodes[-1] and !str2bool($::ceph_conf) {
+    class { 'ceph::deploy':
+      auth_supported   => 'cephx',
+      osd_journal_size => '2048',
```
xdel: should be configurable via user parameter in the parent manifest

deployment/puppet/ceph/manifests/cinder.pp
```diff
@@ -0,0 +1,52 @@
...
+  $volume_driver,
+  $rbd_pool,
+  $glance_api_version,
+  $rbd_user,
+  $rbd_secret_uuid,
+) {
+  if str2bool($::cinder_conf) {
+
+    exec {'Copy configs':
+      command => "scp -r ${mon_nodes[-1]}:/etc/ceph/* /etc/ceph/",
```
xdel, line 11: Do you think this is really needed? Ceph-deploy should take care of this.

deployment/puppet/ceph/manifests/nova_compute.pp
```diff
@@ -0,0 +1,26 @@
...
+    exec {'Copy conf':
+      command => "scp -r ${ceph_nodes[-1]}:/etc/ceph/* /etc/ceph/",
+      require => Package['ceph'],
+      returns => [0,1],
+    }
+    file { '/tmp/secret.xml':
+      content => template('ceph/secret.erb')
+    }
+    exec { 'Set value':
+      command => 'virsh secret-set-value --secret $(virsh secret-define --file /tmp/secret.xml | egrep -o "[0-9a-fA-F]{8}(-[0-9a-fA-F]{4}){3}-[0-9a-fA-F]{12}") --base64 $(ceph auth get-key client.volumes) && rm /tmp/secret.xml',
```
xdel, line 15: Holy cow!

deployment/puppet/ceph/manifests/ssh.pp
```diff
@@ -0,0 +1,40 @@
...
+  }
+  file { '/root/.ssh/id_rsa':
+    source => $ssh_private_key,
+  }
+  file { '/root/.ssh/id_rsa.pub':
+    source => $ssh_public_key,
+  }
+  file { '/etc/ssh/ssh_config':
+    mode    => '0600',
+    content => "Host *\n  StrictHostKeyChecking no\n  UserKnownHostsFile=/dev/null\n",
```
xdel, line 38: This file should be replaced by our ssh magic soon.

plechenko:
add to whitelist

deployment/puppet/ceph/manifests/cinder.pp
xarses, line 11: hold over from original script. This wouldn't be needed for nodes with osd or mon roles. I think the use case was for mon/osd not being on the node that cinder and glance where being configured on. Assuming mon's on controllers and osd's on storage nodes; this would be necessary for the compute nodes.

deployment/puppet/ceph/manifests/ssh.pp
xarses, line 38: yes it should =)

angdraug:
Test failed due to github error, please retest.

plechenko:
retest this please

deployment/puppet/ceph/manifests/nova_compute.pp
xarses, line 15: other than the egrep this is from the setup docs http://ceph.com/docs/next/rbd/rbd-openstack/

xdel, line 15: I meant putting ceph key in the tmp without mktemp or something alike looks awful. Predefined secret looks very bad too(you have get-or-create for such situation).

deployment/puppet/ceph/manifests/deploy.pp
xdel: Oh I said computes.. cinders ofc. Please think about 50 pgs/node with assumption we`ll have two or three pools with heavy amounts of data in observable future.

deployment/puppet/ceph/manifests/cinder.pp
xdel, line 11: Hmm.. Ceph-deploy will spread config from ''initial'' node everywhere(or it needs to :) ), so I`m in doubt if such action needed anyway.

deployment/puppet/ceph/manifests/deploy.pp
```diff
@@ -0,0 +1,89 @@
...
+    target => "/etc/ceph/ceph.client.admin.keyring",
+  }
+  class p_osd {
+    define int {
+      $devices = join(suffix($osd_nodes, ":${name}"), " ")
+      exec { "preparing drives`s on ${devices}":
+        command => "ceph-deploy osd prepare ${devices}",
+        returns => 0,
+        timeout => 0, #TODO: make this something reasonable
+        tries => 60,  #This is necessary because of race for mon creating keys
```
xdel: What is this thing?

deployment/puppet/ceph/manifests/deps.pp
```diff
@@ -0,0 +1,12 @@
+class ceph::deps (
+  $type = 'base'
+){
+  
+  package { ['ceph', 'redhat-lsb-core','ceph-deploy', 'python-pushy']:
```
xdel, line 5: Let`s install everything including rgw, it may shorten our job during future upgrades.

deployment/puppet/ceph/manifests/ssh.pp
```diff
@@ -0,0 +1,40 @@
+class ceph::ssh{
+  package {['openssh-server', 'openssh-clients']:
+    #TODO: debian == openssh-client
+    #TODO: rhel == openssh-clients
+    ensure => latest
+  }
+  $ssh_private_key = 'puppet:///modules/ceph/openstack'
+  $ssh_public_key = 'puppet:///modules/ceph/openstack.pub'
+
+  File {
```
xdel, line 10: What file?

deployment/puppet/ceph/manifests/deps.pp
xarses, line 5: rgw support is coming in to a push near you ... soon.

deployment/puppet/ceph/manifests/deploy.pp
xarses: ceph-deploy osd prepare is ensuring there is a filesystem on the disk according to the args passed to ceph.conf (above). 

timeout: It has a long timeout because of the format taking forever. A resonable amount of time would be around 300 times the length of $osd_nodes. Right now its 0 to prevent puppet from aborting it.

tries: there is a race condition that this is preventing, after the 'ceph-deploy create mon' the bootstrap and admin keys must exist before the osd commands can be issued. This however requires the mon node to establish a quorum prior which takes 10-15 seconds for me. because 'ceph-deploy create mon' or 'ceph-deploy gather-keys' doesn't block, this seemed like an OK place to block, but it really should be in with 'ceph-deploy gather-keys'

deployment/puppet/ceph/manifests/cinder.pp
xarses, line 11: the current logic (which i want to re-factor) is to setup each of the mon's then run ceph-deploy on the last one, so the key's aren't present until mon[-1].

deployment/puppet/ceph/manifests/deploy.pp
xarses: See http://ceph.com/docs/next/rados/operations/pools/#createpool and http://ceph.com/docs/next/rados/operations/placement-groups/ they imply that they shouldn't be changed aftewards so we will need a better way to calculate what we think we want this value to be. At a minimum, this should be changed to a parameter

deployment/puppet/ceph/manifests/ssh.pp
xarses, line 10: because its a capital F, I think its setting resource defaults for the remaining "file" declarations, It's a hold over from the import from the internal repo.

deployment/puppet/ceph/manifests/deploy.pp
angdraug: Usually when code needs explanations that long, it's worth putting the explanations in the comments near the code, so that no one tries to refactor it away later without fully understanding why it was done this way in the first place.

xdel: May be we cay try to rely on the output of ceph-deploy and fail if nothing comes out in relatively large time interval, instead of doing this?

deployment/puppet/ceph/manifests/deploy.pp
xdel: User probably don`t know how to change this parameter properly without such knowledge so at first place there should be an automated variable and we can expose it to the parent in remaining time.

xarses:
rebased  087c323 and 719303e from internal into this branch to add radiosgw support added 3c7d07d to clean up the rebase

deployment/puppet/ceph/manifests/deploy.pp
xarses: ceph-deploy create mon dosen't block for the quorum, we have to wait somewhere. Are you proposing that we monitor the log? In the mean time, I moved this to a new block using ceph-deploy gather-keys instead of 'osd prepare'

xdel:
+1

deployment/puppet/ceph/README.md
```diff
@@ -0,0 +1,307 @@
...
+After installing onto a fuel cluster
+
+1. Define your partitions. If you will re-define any partations you must make
+sure they are exposed in the kernel before running the scripts see ``partx -a
+/dev/<device>`` after ``umount /boot``.
+
+2. Copy ``fuel-pm:/etc/puppet/modules/*`` to ``{node}:/etc/puppet/modules``
+3. Copy ``/etc/puppet/modules/ceph/examples/site.pp`` to ``/root/ceph.pp``.
+4. Edit ceph.pp for desired changes to ``$mon_nodes``, ``$osd_nodes``, and ``$osd_disks``.
+5. Run ``puppet apply ceph.pp`` on each node **except** ``$ceph_nodes[-1]``,
```
bogdando, line 288: What does it mean? Except last one, but "then run the same command on that last one". When then?

angdraug, line 288: It means that `puppet apply ceph.pp` should be run on `$ceph_nodes[-1` only after `puppet apply ceph.pp` has finished on all other nodes. I'll create a patch to clarify the wording here.

PR #504, opened 2013-08-27
Double var declaration

warpc:
Should be strongly apply with Astute pull: https://github.com/Mirantis/astute/pull/61 because https://github.com/Mirantis/astute/pull/61/files#L17R19


dmitryilyin:
+1

PR #509, opened 2013-08-28
added heat, murano, savanna modules.

sirmax123:
Could you please merge this code .

### DIFF
--- a/deployment/puppet/ceph/manifests/deploy.pp
+++ b/deployment/puppet/ceph/manifests/deploy.pp
@@ -123,7 +123,7 @@ class ceph::deploy (
   if $mds_server {
     exec { 'ceph-deploy-s4':
       command => "ceph-deploy mds create ${mds_server}",
-      require => Class['c_osd']
+      require => Class['c_osd'],
       logoutput => true,
     }
   }
@@ -131,7 +131,7 @@ class ceph::deploy (
     define int {
       exec { "Creating pool ${name}":
         command => "ceph osd pool create ${name} ${osd_pool_default_pg_num} ${osd_pool_default_pgp_num}",
-        require => Class['c_osd']
+        require => Class['c_osd'],
         logoutput => true,
       }
     }
--- a/deployment/puppet/mysql/manifests/replicator.pp
+++ b/deployment/puppet/mysql/manifests/replicator.pp
@@ -1,4 +1,4 @@
-class mysql::replicator (
+#class mysql::replicator (
 # Not used because we need to start mysql for the first time 
 # with corosync with init script to create users
 #  database_user { "${user}@${name}":
@@ -10,4 +10,4 @@ class mysql::replicator (
 #    provider => 'mysql',
 #    require => Database_user["${user}@${name}"]
 #  }
-}
+#}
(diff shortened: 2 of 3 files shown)


plechenko:
add to whitelist

dmitryilyin:
- Модуль heat был взят готовым с puppetlabs и его можно считать стандартным
- Модули murano и savanna не используют init и используют только db, api, conductor. Init не нужен?
- Я не вижу нигде установки и настройки плагинов к Горизонту. Это делается rpm пакетами?
- Обновление модуля inifile, keystone и stdlib в fuel нежелательно без веских причин. Возможно будет работать и на старой версии?
- Зачем создано столько типов для работы с ini файлами, отличающиеся только именем файла? Можно было использовать только один и передавать имя файла параметром? Хотя это мелкая проблема...
- heat устанавливается на контроллерах? не лучше ли его установить на compute
- Поддержка quantum сети? savanna? murano?
- Как обстоять дела с запуском служб в HA конфигурации? Запуск экземпляров на всех контроллерах одновременно? Хранят ли сервисы своё состояние только в базе данных или есть промежуточное хранилище?
- Для завершения интеграции нужно будет собрать отдельный манифест установки savanna и манифест установки murano, вставить в них переменные из фактов в naily.facts, сделать в веб интерфейсе опции "Установить savanna" и "Установить murano" и передавать их в факты. При включении этих фактов импортировать соответствующий манифест в site.pp. Кроме этого потребуется размещение на iso дополнительных репозиториев с rpm пакетами для fuel и savanna.
- Сейчас поддерживается только CentOS. Поддержка Ubuntu пока не планируется. Возможно будет работать в RHEL.

# 
- Heat module was taken from puppetlabs and can be considered standard one.
- Savanna and murano modules are not using init parts at all? Inits parts are not needed?
- I don't see Horizon plugins installation and configuration nowhere. Perhaps it's done by rpm packages?
- Updating stdlib, inifile and keystone modules should not be done unless there is a very serious reason to. Perhaps our old versions will work fine too?
- Why there are so many inifile type derevatives with only config file name changed? Perhaps it would be beter to just pass file name as an argument? But it's minor problem anyway...
- We install heat on controllers? I thought it would be better to install them on computes?
- How about quantum network support in murano and savanna projects?
- How about HA configurations? Will murano and savana be compatible with them? Maybe run an instance on every controller? Are these services stateless?
- In order to finish integration we need to do the following: write separate manifests for murano and savanna, insert fact variables from naily.facts, create options 'Install murano' and 'Install savanna' in Web interface and pass these options into facts. When either of these options is enables just include corresponding manifest part into our main manifest file using 'if' condition. We also will require to place all new rpm packages on our iso image.
- Now there is support only for CentOS 6.4. Ubuntu support is not planned. Perfaps these packages will work in RHEL too.

xarses:
why are we are we:
- updating keystone?
- why are there so many changes with stdlib?
- why are there many empty stdlib files

angdraug:
Adding 8K lines in a single commit is bad practice, changing other unrelated modules (inifile, stdlib, keystone as mentioned above) in the same commit is even worse. It looks like some of these modules were simply updated to latest upstream versions from puppetlabs. If so, was it really necessary and why?

At the very least, please add references to your sources for each module that you took verbatim from upstream, so that everyone can tell what you've changed from upstream and which version of upsteam your code is based on.

dmitryilyin:
Yes, we definitly should not update other modules and most likely it's not needed at all. These modules were just copied over during commit.
There are a lot of line because it's pretty much initial commit for these projects and without changes to our modules there are not so many lines at all. Let's calculate... only 1659 lines of new code total!

mihgen:
Folks, only, only, only English please.

PR #510, opened 2013-08-28
Fuel key generation

e0ne:
Generate Fuel UUID for registration from Fuel UI


plechenko:
retest this please

plechenko:
retest this please

xarses:
Do we need to call this "generate_fuel_key"? wouldn't "generate_uuid_key" be more suited if anyone where to re-use this?

e0ne:
May be, but also we could change algorithm of Fuel key generation

deployment/puppet/nailgun/lib/facter/generate_fuel_key.rb
```diff
@@ -0,0 +1,5 @@
+Facter.add("generate_fuel_key") do
```
bogdando, line 1: would it generate new key every time it have been called as $::generate_fuel_key ?

e0ne, line 1: Yes, it will. Do we need to store it anywhere in fuel?

adanin, line 1: Yes. What about to create a subtask about it (UUID generation inside of Fuel lib) and assign it to someone from Fuel lib team?

mihgen:
-1. Every time we re-run puppet for any reason, you will get different uuid. This is simply unacceptable.

I suggest another approach - let's generate uuid during anaconda post-install section. Thus, we can be sure it is generated once.

xarses:
Can we stash these like with the new sshkeys?

mihgen:
Umm how do we do it with the new sshkeys?

xarses:
They are being created and stored in astute. see https://github.com/Mirantis/astute/pull/62 and https://groups.google.com/a/mirantis.com/forum/#!topic/fuel-core-team/J1RkK3J6_0o.

Maybe we can extend the generator to be able to generate and store other things such as UUID. This would also be helpful for where other items are being used as "secrets" and should be regenerated, but only for each cluster.

dmitryilyin:
Wait, what are we going to do here?

You have just created a fact that return new random uuid each time facter runs and write this fact into settings.yaml template.

This is bad idea because uuid will change on every puppet run on master node... ok, we usually don't run puppet again but still this method is absolutely incorrect.

The right way to do it would be to execute 'uuidgen > /etc/fuel-uuid' in anaconda post-install script and then write a fact to get value from this file:

Facter.add('fuel_uuid') do
  setcode do
    begin
      IO.read('/etc/fuel_uuid')
    rescue
      nil
    end
  end
end
~  
Next you can add this uuid into settings template

bogdando:
We can as well use dmidecode -s system-uuid

mihgen:
How uniq is it?

On Wed, Sep 4, 2013 at 4:25 PM, Bogdan Dobrelya notifications@github.comwrote:

> We can as well use dmidecode -s system-uuid
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/Mirantis/fuel/pull/510#issuecomment-23785015
> .

## 

Mike Scherbakov

bogdando:
For virtual cases, at least for single KVM host case, it would be, because it represents the virsh domain name which is always uniq.
For HW and other virtual cases, need to be investigated more precisely.

Here is ruby example for dmi system uuid extraction:
require 'rubygems'
require 'ohai'
@os = Ohai::System.new()
@os.all_plugins
uuid = @os[:dmi][:system][:uuid].strip rescue nil

mihgen:
We don't need UUID of each node. We need only one UUID for master node
installation.

On Thu, Sep 5, 2013 at 3:32 PM, Bogdan Dobrelya notifications@github.comwrote:

> For virtual cases, at least for single KVM host case, it would be, because
> it represents the virsh domain name which is always uniq.
> For HW and other virtual cases, need to be investigated more precisely.
> 
> Here is the example of custom fact for uuid would have been extracted from
> DMI on bootstrap stage:
> https://github.com/bogdando/fuelweb/blob/poc_fence/bin/agent#L281-L284
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/Mirantis/fuel/pull/510#issuecomment-23861279
> .

## 

Mike Scherbakov

bogdando:
Indeed. I believe it is possible to populate master uuid as well

On 09/05/2013 02:39 PM, Mike Scherbakov wrote:

> We don't need UUID of each node. We need only one UUID for master node
> installation.
> 
> On Thu, Sep 5, 2013 at 3:32 PM, Bogdan Dobrelya 
> notifications@github.comwrote:
> 
> > For virtual cases, at least for single KVM host case, it would be, 
> > because
> > it represents the virsh domain name which is always uniq.
> > For HW and other virtual cases, need to be investigated more precisely.
> > 
> > Here is the example of custom fact for uuid would have been 
> > extracted from
> > DMI on bootstrap stage:
> > https://github.com/bogdando/fuelweb/blob/poc_fence/bin/agent#L281-L284
> > 
> > —
> > Reply to this email directly or view it on 
> > GitHubhttps://github.com/Mirantis/fuel/pull/510#issuecomment-23861279
> > .
> 
> ## 
> 
> Mike Scherbakov
> 
> —
> Reply to this email directly or view it on GitHub 
> https://github.com/Mirantis/fuel/pull/510#issuecomment-23861591.

## 

Best regards,
Bogdan Dobrelya,
Researcher TechLead, Mirantis, Inc.
+38 (066) 051 07 53
Skype bogdando_at_yahoo.com
38, Lenina ave.
Kharkov, Ukraine
www.mirantis.com
www.mirantis.ru
bdobrelia@mirantis.com

deployment/puppet/nailgun/templates/settings.yaml.erb
```diff
@@ -10,6 +10,8 @@ DATABASE:
   user: "<%= database_user %>"
   passwd: "<%= database_passwd %>"
 
+FUEL_KEY: "<%= $fuel_key %>"
+
```
bogdando, line 14: typo

e0ne:
So should we use 'system-uuid' or 'uuidgen' to generate Fuel key?

mihgen:
Lets finally use uuidgen in for example post install script on master,
which is run once, storing this value to some place like /etc/fuel_uuid,
and let's close this discussion.

Mike Scherbakov

e0ne:
retest this please

e0ne:
retest this please

mihgen:
-1
Please resubmit this PR to master

PR #513, opened 2013-08-31
Replace tinyproxy to apache

Zipfer:


deployment/puppet/osnailyfacter/manifests/params.pp
```diff
@@ -0,0 +1,13 @@
+class osnailyfacter::params {
+  case $::osfamily {
+    'RedHat': {
+      $apachedir = '/etc/httpd/'
+      $apache_site_dir = "${apachedir}conf.d/"
+    }
+    'Debian': {
+      $apachedir = '/etc/apache2/'
+      $apache_site_dir = "${apachedir}site-enabled/"
```
xdel, line 9: sites-enabled
also you need to put file to the sites-available and make symlink to follow Debian layout

deployment/puppet/osnailyfacter/templates/api_proxy.conf.erb
```diff
@@ -0,0 +1,23 @@
...
+<VirtualHost *:8888>
+  ProxyRequests on
+  ProxyVia On
+  AllowCONNECT 443 563
+  ServerAdmin webmaster@localhost
+  HostnameLookups off
+
+  ErrorLog syslog:local1
+  # Possible values: debug, info, notice, warn, error, crit, alert, emerg
+  LogLevel debug
```
xdel, line 13: Not sure if we want debug for such thing as tests.

deployment/puppet/osnailyfacter/manifests/params.pp
Zipfer, line 9: in puppet case it doesn't need, it should be only with manual apache configuration

xdel, line 9: Not understood this. You`re creating directory outside of default inclusion in apache.conf and saying that it should work. Please double-check my first remark and fix this part.

PR #514, opened 2013-09-02
replace tinyproxy to apache

Zipfer:


deployment/puppet/osnailyfacter/manifests/apache_api_proxy.pp
```diff
@@ -0,0 +1,35 @@
...
+    action  => 'accept',
+    require => Class['openstack::firewall'],
+  }
+
+  if ($::osfamily == 'Debian') {
+
+    file { '/etc/apache2/sites-available/api_proxy.conf':
+      content => template('osnailyfacter/api_proxy.conf.erb'),
+      require => Package['httpd'],
+      notify  => Service['httpd'],
```
xdel: ehrm, this it Puppet - you`ll need to ensure that the both file and symlink exists and then send notify

xarses:
This breaks cluster_ha.pp:
Could not retrieve catalog from remote server: Error 400 on SERVER: Puppet::Parser::AST::Resource failed with error ArgumentError: Could not find declared class osnailyfacter::tinyproxy at /etc/puppet/modules/osnailyfacter/manifests/cluster_ha.pp:187 on node ...

PR #516, opened 2013-09-02
Remove old versions of cobbler manifests for rhel

dmi-try:


deployment/puppet/cobbler/templates/kickstart/rhel.ks.erb
```diff
@@ -1,162 +0,0 @@
-# WHAT TO DO (install fresh system rather than upgrade)
```
xdel, line 1: so there will be only one remaining cobbler kickstart - centos.ks.erb, right?

PR #517, opened 2013-09-02
Apache redirect on admin node

dmi-try:


xdel:
Will close due to near migration to nginx. Is someone having any objections please comment here.

xarses:
Will we have nginx for 3.2? If yes, then I'm OK to see this closed.

PR #528, opened 2013-09-05
added savanna module

sirmax123:


dmitryilyin:
Same code but now without any changes in our modules. Can be accepted already but will require much more work to make use of it.

PR #546, opened 2013-09-12
Move swift root directory to /srv to prevent error

dmitryilyin:
Thre was file not found error because /var/lib/glance did not exist.
Why do we use  /var/lib/glance for swift root anyway? Let's use common location as in other deployment modes.

### DIFF
--- a/deployment/puppet/osnailyfacter/manifests/cluster_ha.pp
+++ b/deployment/puppet/osnailyfacter/manifests/cluster_ha.pp
@@ -57,7 +57,7 @@ else {
 
 if !$swift_partition
 {
-  $swift_partition = '/var/lib/glance/node'
+  $swift_partition = '/srv/node'
 }
 
 


aglarendil:
-1
This problem is related to Ubuntu disk layout configuration. It will be automatically fixed after Ubuntu provisioning is completed. Not sure we should merge it.

monroid:
+1

xarses:
I don't understand why this is being moved. this default should only impact the controllers, they have to have /var/lib/glance (glance requires it) and the there should be a glance VG mounted here.

dmitryilyin:
Then perhaps this error is related to dependencies. For example swift could try to install into /var/lib/glance which have not been created jet...
Or it should already be there when have been provisiond?

This definitly needs more investigation.

xarses:
I'm not sure how the folder dosn't exist since cobbler creates it and mounts the image-glance partition there. 

/dev/mapper/image-glance  5.0G   33M  5.0G   1% /var/lib/glance

I intend to clean up both cluster_simple and cluster_ha and will re-factor an amount of the swift logic so that it can be turned off if proper ceph options are enabled. I can look at the dep tree then.

mihgen:
-1
/var/lib/glance is created by disk layout configurator. If you want to use the other place, make sure to change disk layout configurator. 
This folder must exist prior to puppet run, so we basically should not need this.

PR #550, opened 2013-09-13
[PRD-999] add nodes internal addresses into hosts file for name resoluti...

aglarendil:
...on

### DIFF
--- a/deployment/puppet/l23network/lib/puppet/parser/functions/nodes_to_hosts.rb
+++ b/deployment/puppet/l23network/lib/puppet/parser/functions/nodes_to_hosts.rb
@@ -1,0 +1,21 @@
+#
+# array_or_string_to_array.rb
+#
+
+module Puppet::Parser::Functions
+  newfunction(:nodes_to_hosts, :type => :rvalue, :doc => <<-EOS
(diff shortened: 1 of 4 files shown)


dmitryilyin:
There is a good built it Puppet resource for /etc/hosts content management.

``` ruby
Host {
  ensure => present,
}

host { 'controller01.domain.ltd' :
  ip => '192.168.0.2',
  host_aliases => [ 'c1', 'controller1', 'controller01' ],
  comment => 'The first Fuel controller',
}
```

It doesn't forbid you from manual editing of /etc/hosts too, but will enforce attributes of described hosts without touching unmanaged ones.

You can also list all installed host resources
**puppet resource host**

Get help about host resource type
**puppet describe host**

And get information about specific host
**puppet resource host localhost**

``` ruby
host { 'localhost':
  ensure => 'present',
  ip     => '127.0.0.1',
  target => '/etc/hosts',
}
```

There are not so many things Puppet is good for but this is definitly one of them)

http://docs.puppetlabs.com/references/latest/type.html#host

dmitryilyin:
Great and elegant solution! +1

PR #551, opened 2013-09-13
fixing wrong dependency order

nikit0z:


deployment/puppet/quantum/manifests/server.pp
```diff
@@ -40,8 +40,8 @@
 
   case $::osfamily {
     'Debian': {
-       Quantum_config<||>->Package[$server_package]
```
xdel, line 43: remove entire ifification please

PR #552, opened 2013-09-13
Savanna

sirmax123:


deployment/puppet/savanna/manifests/dashboard.pp
```diff
@@ -0,0 +1,88 @@
...
+
+  } else {
+    $service_ensure = 'stopped'
+    $line_in_settings_py_ensure = 'absent'
+    $package_ensure = 'absent'
+  }
+
+
+
+  define line($file, $line, $ensure = 'present') {
```
aglarendil: This code is a little bit crappy. Look into puppet concat module or file_line resource to do such things. Or create separate python file and import it.

dmitryilyin:
Yes, we all know that Puppet sucks with file editing and there are a lot of hacks to work around its architectural defects. But using sed patching is your last resort and should be avoided if there are better ways aviable.

If you just want to a line to file if this line is not present just use file_line resource from stdlib that is already aviable here.

Snipplets, templates and concats could be good for configuration files, but not for scripts usually.

There is also augeas that has it own... shortcomings.

You can also just create perl/ruby/python script if you need to perform complex modifications and run it with exec. But make shure that you make it idempotent.

This scripts could be implemented as native Puppet resources that could accept parameters and manage their state.

If all above fails or is too much for the job use sed/perl patching execs))

dmitryilyin:
Now everything looks much better. Ready to be accepted.

deployment/puppet/savanna/manifests/dashboard.pp
```diff
@@ -0,0 +1,51 @@
...
+  file_line{'savannaurl'       : path => $local_settings, line => "${savanna_url_string}",                          ensure => $line_in_settings_py_ensure, notify => Service['httpd'], }
+
+  package { 'savanna_dashboard':
+    ensure => $package_ensure,
+    name   => $::savanna::params::savanna_dashboard_package_name,
+    notify => Service['httpd'],
+  }
+
+
+  service { 'httpd':
```
aglarendil, line 44: This service is duplicated.

PR #553, opened 2013-09-13
added murano-dashboard class to murano module

sirmax123:


sirmax123:
$line_in_settings_py_ensure - it is my mistake and can be removed

Does construction 
Package['murano_dashboard'] -> Exec['fix_horizon_config'] -> Service['httpd']
like notify?

Script comes from package and can be run may times because it has build-in check to avoid multiple fixes in config so it is 100%  safe to run it many times.

dmitryilyin:
Ok, accept this already and let's start integrating these modules into out manifests.

deployment/puppet/murano/manifests/dashboard.pp
```diff
@@ -0,0 +1,39 @@
...
+
+  } else {
+    $service_ensure = 'stopped'
+    $package_ensure = 'absent'
+  }
+
+
+  exec { 'fix_horizon_config':
+    command     => "/usr/bin/modify-horizon-config.sh install  /usr/share/openstack-dashboard/openstack_dashboard/settings.py",
+    notify      => Service['httpd'],
```
xdel, line 23: That`s duplicating notification from string 9

dmitryilyin, line 23: Yes, it is. But duplicate notifications are not errors.

PR #561, opened 2013-09-17
Savanna, Murano and Heat integration WIP

dmitryilyin:


deployment/puppet/heat/lib/facter/heat_version_rh.rb
```diff
@@ -0,0 +1,6 @@
+Facter.add("heat_version_rh") do
+  setcode do
+# heat to be installed
+    Facter::Util::Resolution.exec('/usr/bin/yum list | grep heat-common | awk \'{ print $2 }\'')
```
mihgen: How this is going to work for Ubuntu? Also, comment in the code please why we need such a hack.

dmitryilyin: Most likely i'll remove this and replace by manual switcher

dmitryilyin:
These could become issues:
- There was refactoring in rabbitmq config files. Changes should be tested in simple and ha modes.
- Ubuntu support have not been tested and most likely will requitre some work.
- HA mode should be manually tested to check how does controller fallover work.
- There should be system tests to find out does savanna and muraqno work on real task. This should be tested by running hadoop and wondows deployments on real hardware.

dmitryilyin:
TODO:
- Murano-api should use mysql database instead of sqlite. Packege with mysql support is not yet ready. It's _blocker_.
- Status of OSTF tests should be found out

aglarendil:
+1

PR #562, opened 2013-09-17
PRD-1749 Fix release string for Galera package

mattymo:
Fixes PRD-1749 Failed to deploy controllers in HA mode


dmi-try:
+1

dmi-try:
But this version is broken, so it useless to accept

PR #570, opened 2013-09-19
Puppet parser function sanitize_quantum_config()

xenolog:
for cleaning quantum config given from naily facts


aglarendil:
+1

add functions documentation, please

PR #571, opened 2013-09-19
fix shell warning in Fuel syntax check

angdraug:


utils/jenkins/fuel_syntax_check.sh
```diff
@@ -7,11 +7,11 @@
 
 #all_files=`find -name "*.pp" -o -name "*.erb" -o -name "*.sh" -o -name "*.rb"`
 
-if [ -z "$@" ]; then
```
aglarendil, line 10: where do you use these positional parameters? why do you make this change?

angdraug, line 10: I use it to check the files I've modified before pushing a commit, so that I don't have to wait for jenkins to recheck the whole repo to find that I've missed a character somewhere. I run it like this:

``` sh
utils/jenkins/fuel_syntax_check.sh `git status --porcelain|awk '//{print $2}'`
```

xarses:
+1

utils/jenkins/fuel_syntax_check.sh
aglarendil, line 10: There is no difference in this case if you use $@ $*. Correct me if I'm not right.

angdraug, line 10: I thought so too, but in fact there is. To wit, bash(1):

> @ Expands  to  the  positional  parameters,  starting from one.  When the expansion occurs within double quotes, each parameter expands to a separate word.  That is, "$@" is equivalent to "$1" "$2"  ... 

So whenever you pass more than one file to it, [ complains about too many parameters.

aglarendil:
+1

PR #576, opened 2013-09-24
replace tabs to space

Zipfer:


mihgen:
-1
I suggest to replace tabs on spaces only for those files that developer modifies.

PR #579, opened 2013-09-25
Fuel manifests refactoring WIP

dmitryilyin:
DON'T ACCEPT!


deployment/puppet/osnailyfacter/manifests/cluster_simple.pp
```diff
@@ -1,346 +1,214 @@
...
-  $primary_mons   = $controller
-  $primary_mon    = $controller[0]['name']
-  $glance_backend = 'ceph'
-  class {'ceph': 
-    primary_mon  => $primary_mon,
-    cluster_node_address => $controller_node_address,
-  }
-} else {
-  $glance_backend = 'file'
-}
```
xarses, line 128: Lines 4 - 128: I don't see any of this added back in anywhere, why was all of it removed?

PR #581, opened 2013-09-26
Additional fix PRD-1664

r7vme:
- this change removes duplicate vlan package definition
  in favor l23network class where it already defined.
- first fix breaks quantum+vlan deployment
  because of duplicate package definition.

### DIFF
--- a/deployment/puppet/quantum/manifests/plugins/ovs.pp
+++ b/deployment/puppet/quantum/manifests/plugins/ovs.pp
@@ -18,7 +18,6 @@ class quantum::plugins::ovs (
   # todo: Remove plugin section, add plugin to server class
 
   include 'quantum::params'
-  include 'l23network::params'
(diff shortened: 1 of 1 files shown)


monroid:
Test with jenkins fuellib_pull.iso#20 -- passed

aglarendil:
Sergey, review please and say if it is needed to accept this into our code

deployment/puppet/quantum/manifests/plugins/ovs.pp
```diff
@@ -100,13 +99,6 @@
         'OVS/network_vlan_ranges':  value => $network_vlan_ranges;
         'OVS/bridge_mappings':      value => $br_map_str;
       }
-
```
xenolog, line 103: -1

current code already has changes, equal this patch.

PR #593, opened 2013-10-02
Sync CEPH working branch part 3

xarses:
This will sync the current CEPH working branch

Works with Mirantis/fuelweb#1611. Now uses checkboxes to enable cinder/glance/object parts.

What works:
- MON
- OSD
- Cinder
- Glance
- Nova

Still WIP:
- rados-gw


deployment/puppet/ceph/manifests/apt.pp
```diff
@@ -19,9 +20,9 @@
     location => "http://ceph.com/debian-${release}/",
   }
   apt::source { 'radosgw-apache2':
-    location => "http://gitbuilder.ceph.com/apache2-deb-precise-x86_64-basic/ref/master/",
+    location => 'http://gitbuilder.ceph.com/apache2-deb-precise-x86_64-basic/ref/master/',
```
mihgen, line 23: Why do we need this? Is deployment still going to work if master node / slaves will not have access to Internet?

xdel, line 23: Just ask CI team, but current Ubuntu does have all necessary Ceph-related packages, so you may drop it.

deployment/puppet/ceph/manifests/nova_compute.pp
```diff
@@ -1,37 +1,23 @@
...
-        hasrestart => true,
-        subscribe  => Exec['Set value']
-      }
-    }
+
+  file {'/root/secret.xml':
+    content => template('ceph/secret.erb')
   }
+
+  exec {'Set Ceph RBD secret for Nova':
```
xdel, line 10: BTW qemu WILL expose key in the cmdline so if we care about so-called OpenStack security we must use key with lower than admin privileges for VM operations. In real world nobody will explore such vuln in the wild because shell access to the any node in the env gives a plenty of other possibilities to raise itself in mean of privileges.

deployment/puppet/ceph/templates/rgw.conf.redhat.erb
```diff
@@ -0,0 +1,21 @@
+<VirtualHost *:<%= @rgw_port %>>
+  ServerName <%= @fqdn %>
+  DocumentRoot /var/www
```
xdel: Nope, please add another hier level down. /var/www/ should not serve only one root alone by its purpose.

deployment/puppet/ceph/templates/rgw.conf.ubuntu.erb
```diff
@@ -1,6 +1,6 @@
 FastCgiExternalServer /var/www/s3gw.fcgi -socket /tmp/radosgw.sock
 
-<VirtualHost *:80>
+<VirtualHost *:<%= @rgw_port %>>
   ServerName <%= @fqdn %>
   DocumentRoot /var/www
```
xdel: and everywhere too, please add /s3/

deployment/puppet/osnailyfacter/manifests/cluster_ha.pp
```diff
@@ -339,12 +369,22 @@
         debug                   => $debug ? { 'true' => true, true => true, default=> false },
         verbose                 => $verbose ? { 'true' => true, true => true, default=> false },
       }
+      if $storage_hash['objects_swift'] {
+      class { 'swift::keystone::auth':
+        password         => $swift_hash[user_password],
+        public_address   => $public_vip,
+        internal_address => $management_vip,
```
xdel: Please make a note in haproxy manifest that under no circumstances apache balancing should be changed to hash or rr mode from active+backup.

xdel:
By the way, please add remark considering ultimate NTP stability of the controllers. It needs throughout checking during QA stage.

xarses:
retest this please

deployment/puppet/ceph/manifests/apt.pp
xarses, line 23: This is only a syntax checker fix for a file we don't actively use. It's only kept for internal use.

xdel:
everything is fine, ++
let`s wait for f-web part approval

aglarendil:
+2

PR #595, opened 2013-10-02
Added installing parted debian installer

kozhukalov:


deployment/puppet/cobbler/templates/scripts/pmanager.py
```diff
@@ -543,6 +543,7 @@ def lv(self):
                                 end_size,
                                 disk["size"]))
                     self.late("hdparm -z /dev/{0}".format(disk["name"]))
+                self.late("dd if=/dev/zero of=/dev/{0}{1} bs=1M count=1024".format(disk["name"], pcount))
```
xdel: +10 seconds for a charity. Please consider to push it down by order of magnitude.

PR #603, opened 2013-10-03
Functions for sanitize and preprocessing  Quantum config.

xenolog:


aglarendil:
+1

deployment/puppet/quantum/lib/puppet/parser/functions/lib/sanitize_bool_in_hash.rb
```diff
@@ -0,0 +1,31 @@
...
+        rv[k] = process_array(v)
+      else
+        rv[k] = v
+      end
+    end
+    return rv
+  end
+  process_hash(cfg)
+end
+# vim: set ts=2 sw=2 et :
```
warpc: perhaps it is not required

deployment/puppet/quantum/lib/puppet/parser/functions/sanitize_quantum_config.rb
```diff
@@ -0,0 +1,401 @@
...
+  rb_file = File.join(File.dirname(__FILE__),'lib','sanitize_bool_in_hash.rb')
+  load rb_file if File.exists?(rb_file) or raise e
+end
+
+class MrntQuantum
+  #class method
+  def self.sanitize_array(aa)
+    rv = []
+    aa.each do |v|
+      rv.insert(-1,  case v.class.to_s
```
warpc: You can use class as Class without transform it to string. Also you can use `<<` instead of `insert(-1, value)`

``` ruby
rv << case v.class
   when Hash  then sanitize_hash(v)
   when Array  then sanitize_array(v)
   else v
end
```

deployment/puppet/quantum/lib/puppet/parser/functions/sanitize_quantum_config.rb
```diff
@@ -0,0 +1,401 @@
...
+          when "Hash"  then sanitize_hash(v)
+          when "Array" then sanitize_array(v)
+          else v
+        end
+      )
+    end
+    return rv
+  end
```
warpc, line 15: Also you can rewrite this method as:

``` ruby
def self.sanitize_array(aa)
  aa.reduce([]) do |rv, v|
    rv << case v.class
        when Hash  then sanitize_hash(v)
        when Array  then sanitize_array(v)
        else v
      end
    end
end
```

deployment/puppet/quantum/lib/puppet/parser/functions/sanitize_quantum_config.rb
```diff
@@ -0,0 +1,401 @@
...
+      # todo: use network_roles
+      return get_management_vip()
+    end
+    return rv
+  end
+
+  # classmethod
+  def self.get_amqp_config(cfg)
+    rv = cfg.clone()
+    case cfg[:hosts].class.to_s()
```
warpc, line 57: See comment above

deployment/puppet/quantum/lib/puppet/parser/functions/sanitize_quantum_config.rb
```diff
@@ -0,0 +1,401 @@
...
+    $config = sanitize_quantum_config(parse_json($quantum_json_config))
+
+    EOS
+  ) do |argv|
+  Puppet::Parser::Functions.autoloader.loadall
+  given_config = MrntQuantum.sanitize_hash(sanitize_bool_in_hash(argv[0]))
+  q_conf = MrntQuantum.new(self, given_config)
+  rv = q_conf.generate_config()
+  # pUPPET not allow hashes with SYM keys. normalize keys
+  rv = JSON.load(rv.to_json)
```
warpc: ?

aglarendil:
+2

PR #611, opened 2013-10-04
[PRD-1946] fix wrong variable interpolation for nailgun api url

aglarendil:


mattymo:
+1 I admire the way you can put brackets in the right order.

PR #612, opened 2013-10-04
forgotten directive

xdel:


xdel:
++, forgot iface directive

PR #613, opened 2013-10-04
move delimiter due to cobbler parser bug

xdel:


xdel:
++, broke automated ubuntu key placement

PR #614, opened 2013-10-04
Cinder fixes

dmitryilyin:


deployment/puppet/osnailyfacter/manifests/cluster_ha.pp
```diff
@@ -446,6 +446,12 @@
       package { 'python-amqp':
         ensure => present
       }
+      $roles = node_roles($nodes_hash, $::fuel_settings['id'])
+      if member($roles, 'controller') {
```
aglarendil: controller or primary-controller

aglarendil:
+1

xarses:
+1
We took out the openstack::compute cinder ref in #593. Good catch in swift, didn't see that. 

We also changed $manage_volume so that it works more like $glance_backend, but we really need to be able to outright support multiple backends thoughts?

PR #617, opened 2013-10-04
Ubuntu Ceph-related disk changes

rmoe:
Change all parted commands to run in-target because parted doesn't exist in the Busybox environment.
Label file systems so Ceph can be configured.

### DIFF
--- a/deployment/puppet/cobbler/templates/scripts/pmanager.py
+++ b/deployment/puppet/cobbler/templates/scripts/pmanager.py
@@ -426,6 +426,13 @@ class PreseedPManager(object):
             return self._early.append(command)
         return self._early
 
+    def _getlabel(self, label):
+        if not label:
+            return ""
+        # XFS will refuse to format a partition if the
+        # disk label is > 12 characters.
+        return " -L {0} ".format(label[:12])
+
     def pcount(self, disk_id, increment=0):
         if ((self._pcount.get(disk_id, 0) == 0 and increment == 1) or
             (self._pcount.get(disk_id, 0) >= 5)):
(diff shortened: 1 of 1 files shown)


angdraug:
Merged into #593.

deployment/puppet/cobbler/templates/scripts/pmanager.py
```diff
@@ -489,32 +496,33 @@ def partitions(self):
                 pcount = self.pcount("/dev/%s" % disk["name"], 1)
                 tabmount = part["mount"] if part["mount"] != "swap" else "none"
                 if pcount == 1:
-                    self.late("parted -s /dev/{0} mklabel msdos".format(disk["name"]))
+                    self.late("parted -s /dev/{0} mklabel msdos".format(disk["name"]), True)
```
xdel, line 499: By the way, this should be replaced by nonconditional GPT partitioning. @kozhukalov please comment if you plan to fix this seperately.

xdel:
please fix note above, +1 for the rest

kozhukalov:
I think we do not need to launch 'parted' in chroot environment. We can and actually we should install parted into debian-installer system if we need it. Just use anna-install parted-udeb in partman/early_command.

aglarendil:
guys, we do really need gpt partitions for bigger disks.

kozhukalov:
And we'll do it. But for a start let's make this piece of ... code working with msdos.

xdel:
for which reason? It`s non-root disks case anyway.

aglarendil:
ok, guys, we've finally agreed to use GPT. correct this code, pls.

aglarendil:
Merged in the bigger request. Closing this.

PR #621, opened 2013-10-07
get rid of third-party module

xdel:


xdel:
should be done in same time as our mirror update with newer CI packages

aglarendil:
+1

xdel:
+1, repos are ready

PR #623, opened 2013-10-07
[PRD-1978] strip node- prefix from hostname for mysql serverid

mattymo:
https://mirantis.jira.com/browse/PRD-1978

### DIFF
--- a/deployment/puppet/mysql/manifests/params.pp
+++ b/deployment/puppet/mysql/manifests/params.pp
@@ -16,7 +16,7 @@ class mysql::params {
   $port                = 3306
   $etc_root_password   = false
   $ssl                 = false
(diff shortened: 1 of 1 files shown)


deployment/puppet/mysql/manifests/params.pp
```diff
@@ -16,7 +16,7 @@
   $port                = 3306
   $etc_root_password   = false
   $ssl                 = false
-  $server_id           = delete(delete("$::hostname",'controller-'),'fuel-')
+  $server_id           = delete(delete(delete("$::hostname",'controller-'),'fuel-'),"node-")
```
mihgen, line 19: I believe we can use "uid" option which we pass from Astute.

PR #624, opened 2013-10-07
Murano - added mysql support

sirmax123:


mihgen:
Why close without the reason? Pls always comment.

aglarendil:
It is already included in Dima I. pull request. Closed by his request.

PR #630, opened 2013-10-07
fixed several issues for ubuntu disk partitioning

kozhukalov:


kozhukalov:
Here the following issues are solved
1) disks can be with different interfaces like IDE and SATA
2) we install grub manually on all disks

PR #632, opened 2013-10-07
no local corosync logging

xdel:


xdel:
Root owner on the local log file breaks Ubuntu a little bit in means of crmd/pengine logging.

aglarendil:
+2

PR #634, opened 2013-10-07
Fix scoping issues with site variables breaking deployment

xarses:
fix scoping issues with site.pp and cluster_ha and cluster_simple that cause failed deployments.

### DIFF
--- a/deployment/puppet/osnailyfacter/examples/site.pp
+++ b/deployment/puppet/osnailyfacter/examples/site.pp
@@ -25,7 +25,6 @@ stage {'glance-image':
 
 if $::fuel_settings['nodes'] {
   $nodes_hash = $::fuel_settings['nodes']
-
   $dns_nameservers=$::fuel_settings['dns_nameservers']
   $node = filter_nodes($nodes_hash,'name',$::hostname)
   if empty($node) {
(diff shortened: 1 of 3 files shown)


deployment/puppet/osnailyfacter/manifests/cluster_ha.pp
```diff
@@ -24,7 +24,13 @@
   $cinder_nodes_array = []
 }
 
+# All hash assignment from a dimensional hash must be in the local scope or they will
+#  be undefined (don't move to site.pp)
```
mihgen, line 28: I think we don't pass them from Nailgun for RedHat. What's gonna be in this case?

xarses, line 28: probably put back the tests that site.pp had

xarses:
I put the tests from site.pp into cluster_simple / ha

deployment/puppet/osnailyfacter/manifests/cluster_ha.pp
mihgen, line 28: Thanks for fixing this.
So I'm wondering what happens if we don't pass hash and don't use it? It's gonna be undef, or what? I'm just concerned on such if..else clauses for every such parameter; it's gonna be headache for the long run. In Ruby it would be just nil, and we can easily work with this...

xarses, line 28: if its left undef/nil then puppet will throw a 400 trying to compile the manifest if its accessed as a hash and isn't at least an empty hash. Maybe the sane thing is to have nailgun provide sane defalts, or at least an empty hash.

mihgen:
+1 for now.
But we need to address:
1) duplication of stuff in cluster_simple, cluster_ha
2) if-else clauses for every parameter is gonna get us to the hell in the long term.

aglarendil:
That seems wonderful, but why should we declare the same variables (savanna,murano,heat,etc.) 2 times instead of using the only declaration in site.pp.

deployment/puppet/osnailyfacter/examples/site.pp
```diff
@@ -84,6 +60,8 @@
     $public_int   = $::fuel_settings['public_interface']
     $internal_int = $::fuel_settings['management_interface']
   }
+} else {
```
aglarendil, line 63: NO! NO! GOD! NO! This will break RHOS!

aglarendil:
we did not experience any problems with empty hashes defined in top scope. how were you able to reproduce this? I am reverting this commit back. We should try to reproduce it.

PR #640, opened 2013-10-08
Remove OS restrictions from savanna

dmitryilyin:


xdel:
++, waiting for syntax check

PR #641, opened 2013-10-08
PRD-1716: Fix default route for Centos

bogdando:
Note: https://github.com/Mirantis/fuel/pull/662 depends on this one

ISSUE: Centos affected only.
For system wide configured gateway (/etc/sysconfig/network), there is no ensure 
it would has been applied as a default route. It will be applied only after node's 
reboot or network restart.

SOLUTION: 
We should avoid using "heavy" operations like 'service network restart', thus, 
we should create default route by manually adding it into routing table while 
its config stage (adding GATEWAY to /etc/sysconfig/network) is being executed as well.

We should also ensure right apply order as well (L3_if_downup <||> -> Defaultroute[$def_gateway])

### DIFF
--- a/deployment/puppet/l23network/manifests/l3/defaultroute.pp
+++ b/deployment/puppet/l23network/manifests/l3/defaultroute.pp
@@ -17,6 +17,13 @@ define l23network::l3::defaultroute (
               file  => '/etc/sysconfig/network',
               key   => 'GATEWAY',
               value => $gateway,
+          } ->
+          # FIXME: we should not nuke the system with 'service network restart'...
+          # FIXME: but we should ensure default route will be created somehow
+          exec {'Add default route':
+              path    => '/bin:/usr/bin:/sbin:/usr/sbin',
+              command => "route add default gw ${gateway} || true",
+              unless  => "netstat -r | grep -q 'default.*${gateway}'",
           }
         }
     }
--- a/deployment/puppet/l23network/manifests/l3/defaultroute.pp
+++ b/deployment/puppet/l23network/manifests/l3/defaultroute.pp
@@ -20,9 +20,9 @@ define l23network::l3::defaultroute (
           } ->
           # FIXME: we should not nuke the system with 'service network restart'...
           # FIXME: but we should ensure default route will be created somehow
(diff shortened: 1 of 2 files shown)


deployment/puppet/l23network/manifests/l3/defaultroute.pp
```diff
@@ -17,6 +17,13 @@
               file  => '/etc/sysconfig/network',
               key   => 'GATEWAY',
               value => $gateway,
+          } ->
+          # FIXME: we should not nuke the system with 'service network restart'...
+          # FIXME: but we should ensure default route will be created somehow
+          exec {'Add default route':
```
xenolog: -1

xenolog: reason of prd1716 is not in l23network module

bogdando: It is

bogdando:
tested for centos: 
FLATdhcp, nova-network (HA + simple) all defaults - OK
VLANmanager, nova-network simple (ticket given configuration) - OK
{"release": "3.2", "nailgun_sha": "ae5189c958e8989a9f3860330b4110aaa1c464f6", "ostf_sha": "c7a12a74c35e62338cb063124d88d2a738942883", "astute_sha": "29d422940889e729b15c613cb060cc4d472800bd", "fuellib_sha": "c41ba52ff5371ee5258b6dbede5aeaf408ecc306"}

deployment/puppet/l23network/manifests/l3/defaultroute.pp
```diff
@@ -17,6 +17,13 @@
...
+              path    => '/bin:/usr/bin:/sbin:/usr/sbin',
+              command => "route add default gw ${gateway} || true",
```
bogdando: TOD ip route replace

aglarendil: let's do ip route replace

aglarendil:
+1

xenolog:
+1 if tested on local env

PR #650, opened 2013-10-08
use network settings from astute.yaml in ceph

angdraug:
A couple ceph class parameters were missed in transition to astute.yaml.


xarses:
+1

aglarendil:
+1

PR #658, opened 2013-10-09
Add caching mode; cache=none breaks qemu-1.2 and 0.12 in RH derivatives

xdel:


xdel:
+1

bogdando:
Isn't it only GlusterFS affected by?

xdel:
Based on report from services team.

aglarendil:
+1

PR #662, opened 2013-10-09
PRD-2023: Fix default route for Ubuntu

bogdando:
Note: depends on https://github.com/Mirantis/fuel/pull/641

Assumes never put several interfaces to single L2 broadcast domain, if one of them is dhcp enabled, otherwise:
ISSUE:
For vlan/bond enabled interfaces, multiple or wrong default routes might have been created
on start-up time, thus, preventing desired gateway x.y.z.t would have been
added for default route on its static subinterface(vlan/bond slave) start-up.

F.e. for eth0 auto with dhcp and its eth0.100 subint with gateway x.y.z.t defined:
1) eth0 starts up
2) default route have created from dhcp-defined gateway f.g.h.i via eth0 (or ethZ, if it shares the 
same L2 broadcast domain with eth0)
3) eth0.100 starts up
4) gateway x.y.z.t ignored as well ...

To avoid this, you can add pre-up hooks for every main/slave interface to erase existing
default route (if any) configured via this or any other main interface, but that would be an ugly solution.

---

ISSUE: Ubuntu affected only.
For interface with gateway defined (/etc/network/interfaces.d/ifcfg-fooX.YYY), there is no ensure
it would has been applied as a default route. It will be applied only after node's
reboot or network restart.

SOLUTION:
We should avoid using "heavy" operations like 'service network restart', thus,
we should as well create default route by manually adding it into routing table, after interface'
configuration and down&up operation has been completed.

We should also ensure right apply order as well (L3_if_downup <||> -> Defaultroute[$def_gateway])

### DIFF
--- a/deployment/puppet/l23network/manifests/l3/defaultroute.pp
+++ b/deployment/puppet/l23network/manifests/l3/defaultroute.pp
@@ -9,9 +9,9 @@ define l23network::l3::defaultroute (
 ){
   case $::osfamily {
     /(?i)debian/: {
-        exec {'Add default route':
+        exec {'Default route':
             path    => '/bin:/usr/bin:/sbin:/usr/sbin',
-            command => "route add default gw ${gateway} || true",
+            command => "ip route replace default via ${gateway} || true",
             unless  => "netstat -r | grep -q 'default.*${gateway}'",
         }
     }
--- a/deployment/puppet/l23network/manifests/l3/defaultroute.pp
+++ b/deployment/puppet/l23network/manifests/l3/defaultroute.pp
@@ -9,7 +9,11 @@ define l23network::l3::defaultroute (
 ){
   case $::osfamily {
     /(?i)debian/: {
-        fail("Unsupported for ${::osfamily}/${::operatingsystem}!!! Specify gateway directly for network interface.")
+        exec {'Add default route':
+            path    => '/bin:/usr/bin:/sbin:/usr/sbin',
+            command => "route add default gw ${gateway} || true",
+            unless  => "netstat -r | grep -q 'default.*${gateway}'",
+        }
     }
     /(?i)redhat/: {
         if ! defined(Cfg[$gateway]) {
--- a/deployment/puppet/l23network/templates/ipconfig_Debian_bondslave.erb
+++ b/deployment/puppet/l23network/templates/ipconfig_Debian_bondslave.erb
@@ -1,3 +1,4 @@
 auto <%= interface %>
 iface <%= interface %> inet manual
 bond-master <%= @bond_master %>
+pre-up route del default <%= /^(\w+\d+)\.(\d+)/.match(interface)[1] rescue /^(\w+\d+)/.match(interface)[1] rescue "nullface".to_s %> || true
--- a/deployment/puppet/l23network/templates/ipconfig_Debian_bondslave.erb
+++ b/deployment/puppet/l23network/templates/ipconfig_Debian_bondslave.erb
@@ -1,4 +1,3 @@
 auto <%= interface %>
 iface <%= interface %> inet manual
 bond-master <%= @bond_master %>
-pre-up route del default <%= /^(\w+\d+)\.(\d+)/.match(interface)[1] rescue /^(\w+\d+)/.match(interface)[1] rescue "nullface".to_s %> || true
--- a/deployment/puppet/l23network/templates/ipconfig_Debian_dhcp.erb
+++ b/deployment/puppet/l23network/templates/ipconfig_Debian_dhcp.erb
@@ -8,4 +8,3 @@ bond-mode <%= @bond_mode %><% if @bond_miimon %>
 bond-miimon <%= @bond_miimon %><% end %><% if @bond_lacp_rate %>
 bond-lacp-rate <%= @bond_lacp_rate %><% end %>
 <% end %>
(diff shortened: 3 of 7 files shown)


bogdando:
Don't merge! Its under test

bogdando:
tested ubuntu simple nova-network FLATdhcp all defaults
{"release": "3.2", "nailgun_sha": "b83cc7952d96c268e04359f530d91b92162d149c", "ostf_sha": "f3957b3759263edad5f48da2fb367bd20cbb78cc", "astute_sha": "c688fa1d5e513a2d65d3fb43b25a1b1e45337cff", "fuellib_sha": "826bf712b70757d107aebeca184b34d7f2428cb9"}

deployment/puppet/l23network/templates/ipconfig_Debian_dhcp.erb
```diff
@@ -7,4 +7,5 @@ iface <%= interface %> inet dhcp
 bond-mode <%= @bond_mode %><% if @bond_miimon %>
 bond-miimon <%= @bond_miimon %><% end %><% if @bond_lacp_rate %>
 bond-lacp-rate <%= @bond_lacp_rate %><% end %>
-<% end %>
\ No newline at end of file
+<% end %>
+pre-up route del default <%= /^(\w+\d+)\.(\d+)/.match(interface)[1] rescue /^(\w+\d+)/.match(interface)[1] rescue "nullface".to_s %> || true
```
xenolog: line11 may break network configuration.

-1

deployment/puppet/l23network/templates/ipconfig_Debian_manual.erb
```diff
@@ -8,4 +8,5 @@ down ip l set <%= interface %> down
 bond-mode <%= @bond_mode %><% if @bond_miimon %>
 bond-miimon <%= @bond_miimon %><% end %><% if @bond_lacp_rate %>
 bond-lacp-rate <%= @bond_lacp_rate %><% end %>
-<% end %>
\ No newline at end of file
+<% end %>
+pre-up route del default <%= /^(\w+\d+)\.(\d+)/.match(interface)[1] rescue /^(\w+\d+)/.match(interface)[1] rescue "nullface".to_s %> || true
```
xenolog: line12 may break network configuration.

-1

deployment/puppet/l23network/templates/ipconfig_Debian_static.erb
```diff
@@ -13,6 +13,7 @@ bond-mode <%= @bond_mode %><% if @bond_miimon %>
 bond-miimon <%= @bond_miimon %><% end %><% if @bond_lacp_rate %>
 bond-lacp-rate <%= @bond_lacp_rate %><% end %>
 <% end %>
+pre-up route del default <%= /^(\w+\d+)\.(\d+)/.match(interface)[1] rescue /^(\w+\d+)/.match(interface)[1] rescue "nullface".to_s %> || true
```
xenolog: line16 may break network configuration.

deployment/puppet/l23network/manifests/l3/defaultroute.pp
```diff
@@ -9,7 +9,11 @@
 ){
   case $::osfamily {
     /(?i)debian/: {
-        fail("Unsupported for ${::osfamily}/${::operatingsystem}!!! Specify gateway directly for network interface.")
+        exec {'Add default route':
+            path    => '/bin:/usr/bin:/sbin:/usr/sbin',
+            command => "route add default gw ${gateway} || true",
```
bogdando: TODO ip route replace

deployment/puppet/l23network/templates/ipconfig_Debian_static.erb
bogdando: OK, will remove it, suggesting the fact "never put several interfaces to single L2 broadcast domain, if one of them is dhcp enabled" would be always TRUE

deployment/puppet/l23network/templates/ipconfig_Debian_manual.erb
bogdando: OK, will remove it, suggesting the fact "never put several interfaces to single L2 broadcast domain, if one of them is dhcp enabled" would be always TRUE

deployment/puppet/l23network/templates/ipconfig_Debian_dhcp.erb
bogdando: OK, will remove it, suggesting the fact "never put several interfaces to single L2 broadcast domain, if one of them is dhcp enabled" would be always TRUE

deployment/puppet/l23network/templates/ipconfig_Debian_bondslave.erb
```diff
@@ -1,3 +1,4 @@
 auto <%= interface %>
 iface <%= interface %> inet manual
-bond-master <%= @bond_master %>
\ No newline at end of file
+bond-master <%= @bond_master %>
+pre-up route del default <%= /^(\w+\d+)\.(\d+)/.match(interface)[1] rescue /^(\w+\d+)/.match(interface)[1] rescue "nullface".to_s %> || true
```
bogdando: Will remove it, suggesting the fact "never put several interfaces to single L2 broadcast domain, if one of them is dhcp enabled" would be always TRUE

xenolog:
+1

aglarendil:
+1

PR #666, opened 2013-10-09
Return murano rabbit

dmitryilyin:


mihgen:
Are we going to accept this? If so, please provide how you tested this.

PR #668, opened 2013-10-09
drop objects_swift storage backend setting

angdraug:
Enable Swift if Ceph isn't selected as both images and object storage
API backend.


aglarendil:
+2

PR #670, opened 2013-10-09
Ubuntu CEPH fixes

xarses:
Fixes some errors found in the ceph deployment.
- fix error in glance::backend:ceph using wrong glance-api name
- fix mod_ssl package name

### DIFF
--- a/deployment/puppet/ceph/manifests/params.pp
+++ b/deployment/puppet/ceph/manifests/params.pp
@@ -44,7 +44,7 @@ class ceph::params {
       $service_radosgw            = 'radosgw'
       $package_radosgw            = 'radosgw'
       $package_fastcgi            = 'libapache2-mod-fastcgi'
-      $package_modssl             = ''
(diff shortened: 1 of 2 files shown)


deployment/puppet/ceph/manifests/params.pp
```diff
@@ -44,7 +44,7 @@
       $service_radosgw            = 'radosgw'
       $package_radosgw            = 'radosgw'
       $package_fastcgi            = 'libapache2-mod-fastcgi'
-      $package_modssl             = ''
+      $package_modssl             = undef
       $dir_httpd_conf             = '/etc/httpd/conf/'
       $dir_httpd_sites            = '/etc/apache2/sites-available/'
```
xdel, line 49: Please mind about symlink creation. Otherwise +1

mihgen:
+1. Does not break CentOS deployment of controller,compute,cinder - tested on 129th ISO

PR #675, opened 2013-10-10
RHOS needs qemu-kvm-rhev instead of qemu-kvm

mattymo:


mihgen:
What does it fix? Why do we need this change?

aglarendil:
+2

PR #677, opened 2013-10-10
Quantum fix: Wrong calculate local ip address for GRE

xenolog:
- remove debug print


xenolog:
tested on ISO-132, quantum, simple, GRE

aglarendil:
+2

PR #681, opened 2013-10-10
trivial fix

xdel:


mihgen:
+1
Fixes PRD-2105
Applied on top of 138th ISO, puppet continued to run.

PR #682, opened 2013-10-11
Ceph radosgw fixes FOR TESTING - DO NOT MERGE

xarses:
RadosGW requires keystone PKI token format, this would make keystone use PKI token_format when using ceph.

This requires https://mirantis.jira.com/browse/OSCI-363 for CentOS

This has not been tested, this is for testing and review


xarses:
retest this please

PR #685, opened 2013-10-11
place Ceph admin keyring in /etc/ceph

angdraug:
This change fixes Nova failure to attach a Ceph-backed volume to a VM.

Tested with CentOS simple deployment: deployed, launched instance from the TestVM Cirros image, created and attached volume to the instance.


xarses:
+1

xdel:
+1

PR #691, opened 2013-10-12
FIX: parameter name for DEFAULT/quantum_admin_username in nova.conf

xenolog:
- warning


xenolog:
tested centos/simple/quantum/gre

PR #693, opened 2013-10-14
FIX: Calculate segment_id for VLAN mode

xenolog:
PRD-2130


xenolog:
tested on centos/simple/quantum/vlan
tested on centos/simple/quantum/gre

aglarendil:
+2

PR #694, opened 2013-10-14
FIX: private_interface not always present in config

xenolog:
PRD-2172


xenolog:
tested on centos/simple/quantum/vlan
tested on centos/simple/quantum/gre

aglarendil:
+2

PR #698, opened 2013-10-15
[PRD-1552] CEPH RadosGW support

xarses:
Adds working RadosGW (rgw) support

RadosGW to keystone integration requires keystone token_format = PKI

PKI token format is currently broken in Python 2.6 distros. When osfamily == RedHat, UUID format will be forced and keystone integration will be disabled. RadosGW users will have to be configured by hand using radosgw-admin

adds ability to define token_format at top in site and osnailly manifests.
fixes pki_setup exec in keystone

Testing status: (iso 157)
centos-simple  - pass
centos-ha        - pass
ubuntu-simple  - overall fail
                      + glance ceph - fail (missing package python-ceph)
                      + cinder ceph - pass
                      + PKI - pass
                      + ceph mon - pass
                      + ceph osd - pass
                      + radosgw - pass
                      + object in horizon - pass 
ubuntu-ha        - untested

### DIFF
--- a/deployment/puppet/ceph/manifests/init.pp
+++ b/deployment/puppet/ceph/manifests/init.pp
@@ -26,7 +26,7 @@ class ceph (
       $rgw_keyring_path                 = '/etc/ceph/keyring.radosgw.gateway',
       $rgw_socket_path                  = '/tmp/radosgw.sock',
       $rgw_log_file                     = '/var/log/ceph/radosgw.log',
-      $rgw_use_keystone                 = $::token_format ? { 'PKI'=>true, default=>false },
+      $rgw_use_keystone                 = false,
       $rgw_keystone_url                 = "${cluster_node_address}:5000",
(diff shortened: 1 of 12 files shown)


aglarendil:
PKI is still in our release plans, but we are not going to support this token format in 3.2 release. This will not work in HA mode, cause we need confident support for keys distribution among controllers.

aglarendil:
Is there any way to revert to UUID token format for radosgw ?

xarses:
as far as i can tell, test, pry out of the code and devs, RadosGW will only interact with keystone using PKI

xarses:
e3eddcb makes it so token_format is always 'UUID' but allows it to be easily changed if an Advanced user desires to. Since token_format is usually going to be 'UUID' radosgw will not be configured to integrate with keystone. In this case, Swift will be installed anyway. The ceph for object option will be dropped soon.

xarses:
closing this another way was found

PR #702, opened 2013-10-15
FIX: Quantum-server in HA mode respond not on right IP.

xenolog:
Prd 2214


aglarendil:
+2

aglarendil:
Syntax check made manually

PR #703, opened 2013-10-16
[PRD-1552] CEPH RadosGW fixes

xarses:
Many fixes to broken RadosGW feature.
- fix broken order in Ubuntu
- fix missing apache modules in Ubuntu
- fix missing apache rgw.conf in Ubuntu
- fix broken fastcgi.conf in CentOS 
- **_works with UUID token format**_ 
- works with PKI token format

Note: Rados will now use UUID token format by default. This will result in some log messages in /var/log/ceph/rados.log having some error messages about "revoked tokens response is missing signed section" which can be ignored.

Testing status: 
CentOS simple - pass
Ubuntu simple - pass
CentOS ha - ok


xdel:
looks good

angdraug:
+1

xarses:
tests have passed on CentOS and Ubuntu Simple.

PR #709, opened 2013-10-16
Fix missing CEPH monitors in HA

xarses:
Fixes bug introduced by 0fdb733 where HA controllers after first would not create CEPH monitors because unless test passed in wrong condition.

Testing status:
CentOS HA - Pass

OK to merge.


angdraug:
+1

PR #710, opened 2013-10-17
do not merge, only for test

xenolog:


monroid:
PASSED on simple cluster fuellib_pull.iso #29

PR #711, opened 2013-10-17
added bytes per inode for ext3/ext4

kozhukalov:


kozhukalov:
That is about PRD-1046

PR #713, opened 2013-10-17
Add cobbler snippet ubuntu_static_net for Centos and use it instead of dhclient_ignore_routers_opt.

yoojinl:


xdel:
Please put script in the file and execute it as it was done with every other snippet

aglarendil:
+2 should be accepted as soon as https://review.openstack.org/#/c/52165/ is merged into fuel-web

bogdando:
+1

PR #716, opened 2013-10-17
Correct RadosGW endpoint in HA

xarses:
This will point swift endpoint to first controller because there is no VIP in HA for RadosGW and wont be ready in time for 3.2. This will need to be a known limitation in 3.2

Testing status:
centos HA - hand re-run pass
ubuntu HA - testing


angdraug:
+1

aglarendil:
why aren't you using virtual IP for rgw endpoint ?

xarses:
We haven't been able to set one up due to time and testing builds.

PR #723, opened 2013-10-18
Add python-neutronclient to ostf virtualenv

dshulyak:
Currently we are using this hack to install eggs
into ostf virtualenv on master node


dshulyak:
Don't merge it. Need to verify that egg is on every server

dshulyak:
neutronclient doesn't work with http_proxy

PR #727, opened 2013-10-19
[PRD-2323] Fix cinder-volume ceph init on ubuntu

xarses:
- distorize CEPH_ARGS in cinder::volume::ceph
- convert parameters to cinder::params
- remove un-used params from ceph::params
- removes some not needed parts left over from 0fdb733

Resolves: PRD-2323

Testing status: 
Centos HA - pass
Ubuntu - testing


aglarendil:
+2

PR #728, opened 2013-10-19
Fix firewall for radosgw

xarses:
Add missing firewall rule which can cause radosgw to not receive requests

Testing status:
centos ha - pass


aglarendil:
+2

PR #731, opened 2013-10-19
Create flag for don't configure network twice.

xenolog:


xenolog:
tested by hands

xenolog:
L23network is idempotent.

PR #733, opened 2013-10-19
Nailgun NTP improvements

mattymo:


deployment/puppet/cobbler/templates/dhcp-enter-hooks.erb
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
aglarendil, line 1: how is this script related to ntp improvements?

mattymo, line 1: We need dns resolution available during first boot of master node for Fuel Menu and for NTP. It fixes the issues related to time jumps in deployment, which makes following logs difficult. We will enable resolution at boot, then puppet will block dhclient from updating /etc/resolv.conf for DHCP config and set /etc/resolv.conf manually

teran:
Verified, works fine.

aglarendil:
+1

PR #739, opened 2013-10-20
do not create cinder-volume init file on Ubuntu

angdraug:
If init file is created before the package is installed, the package
puts the real init file into /etc/init/cinder-volume.conf.dpkg.dist, so
file_line ends up adding a line into an empty file instead of the real
init file.

Fixes: PRD-2323 (cinder-volume won't start)
Testing: Ubuntu: PASS, CentOS: PASS, please review and merge.


aglarendil:
+2

PR #740, opened 2013-10-20
added puppet packages and deps for our custom projects

mattymo:
Extra deps needed for puppet have to be brought from fuel centos repo.
Murano and savanna are going to be copied to repo, but still are already disabled in UI. It will be necessary to do this later anyway for testing


aglarendil:
+2

PR #743, opened 2013-10-20
[PRD-2360] Radosgw wont start on multiple controllers

xarses:
refactor radosgw key creation so that its idempotent.

testing status: 
tested on ubuntu HA - pass

### DIFF
--- a/deployment/puppet/ceph/manifests/radosgw.pp
+++ b/deployment/puppet/ceph/manifests/radosgw.pp
@@ -185,24 +185,16 @@ class ceph::radosgw (
     mode    => '0755',
     }
 
-  exec { "ceph-create-radosgw-keyring-on ${name}":
-    command => "ceph-authtool --create-keyring ${keyring_path}",
-    creates => $keyring_path,
+  exec { "ceph create ${radosgw_auth_key}":
+    command => "ceph auth get-or-create ${radosgw_auth_key} osd 'allow rwx' mon 'allow rw'",
+  }
+
+  exec { "Populate ${radosgw_auth_key} keyring":
+    command => "ceph auth get-or-create ${radosgw_auth_key} > ${keyring_path}",
+    creates => $keyring_path
   }
 
   file { $keyring_path: mode => '0640', }
-
-  exec { "ceph-generate-key-on ${name}":
-    command => "ceph-authtool ${keyring_path} -n ${radosgw_auth_key} --gen-key",
-  }
-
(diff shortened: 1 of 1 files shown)


angdraug:
+1

deployment/puppet/ceph/manifests/radosgw.pp
```diff
@@ -185,24 +185,16 @@
     mode    => '0755',
     }
 
-  exec { "ceph-create-radosgw-keyring-on ${name}":
-    command => "ceph-authtool --create-keyring ${keyring_path}",
-    creates => $keyring_path,
+  exec { "ceph create ${radosgw_auth_key}":
+    command => "ceph auth get-or-create ${radosgw_auth_key} osd 'allow rwx' mon 'allow rw'",
```
xdel, line 189: For which reason we`re giving rw on mons?

deployment/puppet/ceph/manifests/radosgw.pp
```diff
@@ -185,24 +185,16 @@
...
-    command => "ceph-authtool -n ${radosgw_auth_key} --cap osd 'allow rwx' --cap mon 'allow rw' ${keyring_path}",
+  exec { "Populate ${radosgw_auth_key} keyring":
+    command => "ceph auth get-or-create ${radosgw_auth_key} > ${keyring_path}",
+    creates => $keyring_path
   }
 
-  exec { "ceph-add-to-ceph-keyring-entries-on ${name}":
-    command => "ceph -k /etc/ceph/ceph.client.admin.keyring auth add ${radosgw_auth_key} -i ${keyring_path}",
-  }
+  file { $keyring_path: mode => '0640', }
```
xdel, line 197: again, permission question

deployment/puppet/ceph/manifests/radosgw.pp
angdraug, line 189: This manifest had 'rw' here from the very first version. If you believe we should lock it down please raise a separate ticket.

deployment/puppet/ceph/manifests/radosgw.pp
angdraug, line 197: This line was moved, not added in this commit.

xdel:
+1
permissions on mon should be lowered in the seperate pull

angdraug:
For future reference, here's the reason why mon needs rw for radosgw:
http://ceph.com/docs/master/radosgw/config/#create-default-pools

Once the radosgw pools are created, the mon privileges can be reduced to r.

PR #744, opened 2013-10-21
Allow verbose=true from site.pp

bogdando:
Prevent hardcoded verbose option from being
overrided by False.s

Fixes: #PRD-2348


bogdando:
duplicates https://github.com/Mirantis/fuel/pull/725

PR #750, opened 2013-10-21
Ubuntu install full vim by default

xarses:
Add full vim package to Ubuntu installs, the current vi package provides very bad user experience. the vim package is already in the repo.


angdraug:
+1

xenolog:
+1 +1 +1

PR #751, opened 2013-10-22
[PRD-2426] add biosdevname=0 to deployment

mattymo:


bogdando:
+1 
centos simple, iso 233 node provisioning OK
grep -o biosdevname=0 /proc/cmdline OK

mihgen:
+1

PR #756, opened 2013-10-24
Add support for older cciss devices

xarses:
add detection for if devname is cciss and insert seperator if necessary

### DIFF
--- a/deployment/puppet/cobbler/templates/scripts/pmanager.py
+++ b/deployment/puppet/cobbler/templates/scripts/pmanager.py
@@ -204,7 +204,8 @@ class PManager(object):
                     if part["mount"] != "swap" and tabfstype != "none":
                         disk_label = self._getlabel(part.get('disk_label'))
                         self.post("mkfs.{0} -f $(readlink -f /dev/{1})"
-                                  "{2} {3}".format(tabfstype, disk["id"],
+                                  "{2}{3} {4}".format(tabfstype, disk["id"],
+                                                   self._pseparator(disk["id"]),
                                                    pcount, disk_label))
                         if part["mount"] != "none":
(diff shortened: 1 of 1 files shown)


rmoe:
I don't have HP hardware to test the cciss changes on. I did test both a CentOS and Ubuntu deployment on KVM with these changes and didn't see any issues.

aglarendil:
Vladimir, please clarify. It looks like we do already have similar code in master, don't we ?

kozhukalov:
No, we don't. I sent pull request with the same functionality a couple of weeks ago but then removed it due to duplication. Closed one is https://github.com/Mirantis/fuel/pull/762

deployment/puppet/cobbler/templates/scripts/pmanager.py
```diff
@@ -22,6 +22,12 @@ def __init__(self, data):
         self._rcount = 0
         self._pvcount = 0
 
+    def _pseparator(self, devname):
+        pseparator=""
+        if devname.find("cciss")>0:
```
kozhukalov: Please use space symbols according to pep8. I know that this whole script does not satisfy pep8 requirements but it looks awful without space symbols :-)

kozhukalov:
+1

aglarendil:
was it tested for any regressions?

xarses:
we've been running with this patch applied during the ceph osd changes with out issue.

angdraug:
Whitespace added, double quotes changed to single.

angdraug:
US team had this fix included in our ISO images as soon as this pull request was created, and never seen any related breakage during 2 weeks we were testing it. It is also confirmed to work for the user who originally reported this bug.

angdraug:
Fixed merge conflict with #773.

aglarendil:
+2

PR #759, opened 2013-10-25
new options for Havana

xenolog:


angdraug:
Do we really want to maintain a separate branch for Havana and deal with all the fallout of merging a large feature branch into master at some later point? I think it would be safer to implement Havana-specific changes in a backwards compatible manner and merge them one at a time. That way, everyone is exposed to these changes at the same time and any breakage is detected early and can be fixed quickly and with minimum disruption.

xenolog:
YES. I have some reasons for it:
- quantum renamed to neutron
- some packages requirements different with grizzly
- arrived  new options and deprecated some old.

dmitryilyin:
+1

xdel:
+1

PR #761, opened 2013-10-29
Horizon's fixes for Havana

xenolog:
and some else

### DIFF
--- a/deployment/puppet/horizon/manifests/init.pp
+++ b/deployment/puppet/horizon/manifests/init.pp
@@ -65,7 +65,7 @@ class horizon(
     if ! defined(Package[$name]){
       @package { $name : }
     }
-  } 
+  }
 
   File {
     require => Package['dashboard'],
@@ -187,6 +187,18 @@ class horizon(
         ],
         before  => Service['httpd'],
       }
+
+      #todo: may be need fix
+      Package['dashboard'] -> Exec['horizon_compress_styles']
+      Package['dashboard'] ~> Exec['horizon_compress_styles']
+      Package[$::horizon::params::horizon_additional_packages] -> Exec['horizon_compress_styles']
+      exec { 'horizon_compress_styles':
+        path    => '/bin:/usr/bin:/sbin:/usr/sbin',
+        cwd     => '/usr/share/openstack-dashboard',
+        command => 'python manage.py compress',
+        refreshonly => true
+      }
+      Exec['horizon_compress_styles'] ~> Service['httpd']
     }
     'Debian': {
       A2mod {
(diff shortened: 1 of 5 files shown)


deployment/puppet/horizon/manifests/init.pp
```diff
@@ -187,6 +187,18 @@
         ],
         before  => Service['httpd'],
       }
+
+      #todo: may be need fix
+      Package['dashboard'] -> Exec['horizon_compress_styles']
```
dmitryilyin, line 192: Notification implies dependency both -> and ~> are not needed. Leave only ~>

deployment/puppet/horizon/manifests/init.pp
```diff
@@ -187,6 +187,18 @@
...
+      Package['dashboard'] ~> Exec['horizon_compress_styles']
+      Package[$::horizon::params::horizon_additional_packages] -> Exec['horizon_compress_styles']
```
dmitryilyin, line 194: You don't really need to use leading :: in fully qualified path, but it wouldn't hurt

deployment/puppet/horizon/manifests/init.pp
```diff
@@ -216,10 +228,10 @@
...
     ensure    => 'running',
-    enable    => true,
-    require   => Package["$::horizon::params::http_service", "$::horizon::params::http_modwsgi"],
-    subscribe => File["$::horizon::params::local_settings_path", "$::horizon::params::logdir"]
+    enable    => true
   }
+  File["$::horizon::params::local_settings_path", "$::horizon::params::logdir"] ~> Service['httpd']
+  Package["$::horizon::params::http_service", "$::horizon::params::http_modwsgi"] -> Service['httpd']
 
   if $cache_server_ip =~ /^127\.0\.0\.1/ {
```
dmitryilyin, line 236: why regexp? == '127.0.0.1' ?

deployment/puppet/horizon/manifests/init.pp
xenolog, line 192: I not shure. 
-> give order of implementing resources
~> give refresh one if another change

deployment/puppet/horizon/manifests/init.pp
xenolog, line 194: even in 3.xx ?

deployment/puppet/horizon/manifests/init.pp
xenolog, line 236: string, started as '127.0.0.1'

xdel:
looks good

deployment/puppet/horizon/manifests/init.pp
dmitryilyin, line 194: Yes. $::var means explicit top scope and ignore any lower scope vars
$class::var means out of scope fully qualified lookup. You want this var from that class scope ignoring any global and local matches
$var means local scope -> parent scope -> node scrope -> top scope. first match wins

Difference between 3 and 2 if in what parent scope means
in 2 in it class from what current was inherited and scope where current was declared
in 3 parent means only inheritance and declaration means nothing

For example
$msg = 'top'

class a {
  $msg = "a"
}

class a_child inherits a {
  notify { $msg :}
}
will say 'a' in puppet 2 and 3 both

But
$msg = 'top'

class n1 {
  $msg = 'n1'
  include 'n2'
}

class n2 {
  notify { $msg :}
}

include 'n1'

Will say 'n1' in puppet 2.6, will say 'n1' and deprication warning in 2.7, will say 'top' in puppet3

deployment/puppet/horizon/manifests/init.pp
dmitryilyin, line 192: -> meand only require
~> means require and notify
so ~> is same as -> but stronger

PR #765, opened 2013-10-30
Ceilometer integration WIP

aglarendil:
Do not merge


deployment/puppet/ceilometer/manifests/agent/compute.pp
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  nova_config {
+    'DEFAULT/instance_usage_audit'        : value => 'True';
+    'DEFAULT/instance_usage_audit_period' : value => 'hour';
+  }
+
+  Nova_config<| |> {
+    before +> File_line[
+      'nova-notification-driver-common',
```
aglarendil, line 96: why not simply use nova_config resource ?

deployment/puppet/ceilometer/manifests/agent/compute.pp
```diff
@@ -0,0 +1,113 @@
...
+      'nova-notification-driver-ceilometer'
```
aglarendil, line 97: why not simply use nova_config resource ?

PR #767, opened 2013-10-30
some new parameters for Neutron

xenolog:


deployment/puppet/neutron/manifests/init.pp
```diff
@@ -117,6 +117,7 @@
     'DEFAULT/network_auto_schedule':  value => $neutron_config['L3']['network_auto_schedule'];
     'DEFAULT/router_auto_schedule':   value => $neutron_config['L3']['router_auto_schedule'];
     'DEFAULT/agent_down_time':        value => $neutron_config['server']['agent_down_time'];
+    'agent/root_helper':              value => $neutron_config['root_helper'];
```
xdel, line 120: Just curious when it can be changed.

PR #778, opened 2013-11-01
Сeilometer integration

r7vme:
- HA and simple schemas
- Tested on CentOS 6.4, also it works on Ubuntu with some package dependencies fixes.
- SQL db backend. We've succeded with stress tests on ceilometer with mysql.
- Ceilometer is optional and by default disabled.


deployment/puppet/ceilometer/manifests/agent/compute.pp
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  Nova_config<| |> {
+    before +> File_line[
+      'nova-notification-driver-common',
+      'nova-notification-driver-ceilometer'
+    ],
+  }
+
+  file_line {
```
dmitryilyin: Was there a reason to use File_line instead of Nova_config for 'nova-notification-driver-common' and 'nova-notification-driver-ceilometer' ?

r7vme: it tries to add two equal options "notification_driver" this was cause, but we could refactor this piece in favor notification_driver=driver1,driver2, it also supported and use nova_config here.

dmitryilyin:
Looks very much like some other openstack manifests. And most likely will suffer from the same Puppet containment problem I've described here https://docs.google.com/a/mirantis.com/document/d/1VopEsyBq3SqErX5FI5FwWl02VYqmL6lS4O_yCQ6Air4/edit

For example, this may look like it will force database creation striclty before  other ceilometer classes
Class['openstack::db::mysql'] -> Class['openstack::ceilometer']
But it may not be the case. Dependencies between classes doesn't lead to dependencies between resources inside subclasses. Ad mysql database may or may not try to be created after the start of ceilomemter services.

But there are also a lot of other interdependencies betwen components so this may be already fixed... Connections are too complex for me to untangle without running and making --graph.

Al least expect to meet problems with mysql database/user/permission sometimes not being created when services are started and try to double check this.

deployment/puppet/ceilometer/manifests/agent/compute.pp
dmitryilyin: Well... it's ok if you know what and why you are doing)

monroid:
retest this please

monroid:
retest this please

monroid:
retest this please

monroid:
retest this please

monroid:
retest this please

PR #782, opened 2013-11-05
cpufreq additions

xdel:


xdel:
Depends on https://review.openstack.org/#/c/55526/

PR #785, opened 2013-11-06
FIX: neutron plugin configs order additional fix

xenolog:


xenolog:
tested on centos/simple/GRE

PR #793, opened 2013-11-06
[PRD-2272] disable racy operations on crm services

aglarendil:
Do not merge - more work on needed


xenolog:
Do we need backport it to Havana branch?

angdraug:
Not yet, this patch in its current form isn't sufficient to fix the problem, the work on the fix is still ongoing.

PR #798, opened 2013-11-07
[PRD-2509] turn on quantum security groups

aglarendil:
Turn on quantum secgroups for quantum-enabled installations


xenolog:
+1

xenolog:
tested on centos/simple/VLAN:
- horizon lost Quotas.

PR #806, opened 2013-11-07
[4.0] [PRD-2485] Add support for vlan splinters

xarses:
Add option use_vlan_splinters to site.pp. If enabled and kernelmajversion
 < 3.3 then vlan splinters will be enabled if the driver for an bridge
 interface is present in the devices list in l23network::l2::splinters.

This behavior is disabled by default as not all installs may suffer from this
 and should only be enabled when necessary.

Testing status: testing passed on virtual emulation of e1000

To test, change guest network controller to one of the drivers listed in splinters.pp (ie e1000)

### DIFF
--- a/deployment/puppet/l23network/manifests/l2/port.pp
+++ b/deployment/puppet/l23network/manifests/l2/port.pp
@@ -54,6 +54,12 @@ define l23network::l2::port (
       interface_properties => $interface_properties,
       skip_existing => $skip_existing
     }
+
+    if ($::kernelmajversion =~ /^(2.\d|3.[0-2])/ and $::use_vlan_splinters) {
+      L2_ovs_port[$port] ->
+      l23network::l2::splinters {$port: }
+    }
+
     Service<| title == 'openvswitch-service' |> -> L2_ovs_port[$port]
   }
 }
--- a/deployment/puppet/l23network/manifests/l2/splinters.pp
+++ b/deployment/puppet/l23network/manifests/l2/splinters.pp
@@ -1,0 +1,36 @@
+# == Define: l23network::l2::splinters
+#
+# Splinters will check if the passed interface(name) is known to possibly have
+#  issues and will turn on splinters.
+#
+# This requires that $::use_vlan_splinters is true (defaults to false)
+#
+# ===Parameters
+#
+# [*interface*] (optional) the name of the interface to test, defaults to $name
+
+#
+define l23network::l2::splinters (
+	$interface = $name
+  ) {
+
+  # This is a list of drivers that might need splinters to work with older kernels.
+  # From http://openvswitch.org/cgi-bin/ovsman.cgi?page=utilities%2Fovs-vlan-bug-workaround.8.in
+
+  $drivers = ['8139cp', 'acenic', 'amd8111e', 'atl1c', 'ATL1E', 'atl1', 'atl2',
+              'be2net', 'bna', 'bnx2', 'bnx2x', 'cnic', 'cxgb', 'cxgb3',
+              'e1000', 'e1000e', 'enic', 'forcedeth', 'igb', 'igbvf', 'ixgb',
+              'ixgbe', 'jme', 'ml4x_core', 'ns83820', 'qlge', 'r8169', 'S2IO',
+              'sky2', 'starfire', 'tehuti', 'tg3', 'typhoon', 'via-velocity',
+              'vxge', 'gianfar', 'ehea', 'stmmac', 'vmxnet3']
+  $grep_test = join($drivers, ' -e ')
+
+  exec { "inspect interface and enable splinters on interface ${interface}":
+          command   => "ovs-vsctl set interface ${interface} other-config:enable-vlan-splinters=true",
+          logoutput => on_failure,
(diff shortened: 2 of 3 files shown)


deployment/puppet/l23network/manifests/l2/splinters.pp
```diff
@@ -0,0 +1,30 @@
...
+              'ixgbe', 'jme', 'ml4x_core', 'ns83820', 'qlge', 'r8169', 'S2IO',
+              'sky2', 'starfire', 'tehuti', 'tg3', 'typhoon', 'via-velocity',
+              'vxge', 'gianfar', 'ehea', 'stmmac', 'vmxnet3']
+  $grep-test = join($drivers, ' -e ')
+
+  if (::osfamily == 'RedHat' and ::use_vlan_splinters) {
+    exec { "inspect interface and enable splinters on interface ${interface}":
+            command   => "/usr/bin/ovs-vsctl set interface ${interface} other-config:enable-vlan-splinters=true",
+            logoutput => on_failure,
+            onlyif    => "/sbin/ethtool -i ${interface} | /bin/grep -q ${grep-test}",
```
xdel: remove absolute paths please

xdel:
Looks very promising but requires a bit of attention in the docs too.

deployment/puppet/l23network/manifests/l2/splinters.pp
xarses: Should they be dropped all together, or replaced with path => ?

xdel: Not understood first part of sentence, I mean uncommenting path => and striping out prefixes above. Also we need to be sure that our walkthrough will use only interfaces exposed to the system, otherwise it will produce a lot of output from misguided ethtool.

deployment/puppet/l23network/manifests/l2/splinters.pp
```diff
@@ -0,0 +1,37 @@
...
+              'ixgbe', 'jme', 'ml4x_core', 'ns83820', 'qlge', 'r8169', 'S2IO',
+              'sky2', 'starfire', 'tehuti', 'tg3', 'typhoon', 'via-velocity',
+              'vxge', 'gianfar', 'ehea', 'stmmac', 'vmxnet3']
+  $grep-test = join($drivers, ' -e ')
+
+  if (::osfamily == 'RedHat' and ::use_vlan_splinters) {
+    exec { "inspect interface and enable splinters on interface ${interface}":
+            command   => "ovs-vsctl set interface ${interface} other-config:enable-vlan-splinters=true",
+            logoutput => on_failure,
+            onlyif    => "ethtool -i ${interface} | /bin/grep -q ${grep-test}",
```
xdel: Still no test for ephemeral ovs interfaces. May be if [ -f /sys/class/net/${interface} ]; then ethtool -i ${interface} | /bin/grep -q ${grep-test}; else exit 1; fi

xdel:
+1

xenolog:
which interfaces for we will use this feature?

xdel:
drivers list will match only against corresponding physical ifs, but you need to set this interface as a trunk explicitly, if not, that`s the second part of this work

deployment/puppet/l23network/manifests/l2/port.pp
```diff
@@ -54,6 +54,10 @@
       interface_properties => $interface_properties,
       skip_existing => $skip_existing
     }
+
+    if ($::osfamily == 'RedHat' and $::use_vlan_splinters) {
+      l23network::l2::splinters {$port: }
+    }
```
xenolog, line 61: -1

we do not call it here with this conditions.

I mean we need add parameter in "interface" of network scheme, and allow spliters only if it's allowed

deployment/puppet/l23network/manifests/l2/splinters.pp
```diff
@@ -0,0 +1,37 @@
...
+
+  $drivers = ['8139cp', 'acenic', 'amd8111e', 'atl1c', 'ATL1E', 'atl1', 'atl2',
+              'be2net', 'bna', 'bnx2', 'bnx2x', 'cnic', 'cxgb', 'cxgb3',
+              'e1000', 'e1000e', 'enic', 'forcedeth', 'igb', 'igbvf', 'ixgb',
+              'ixgbe', 'jme', 'ml4x_core', 'ns83820', 'qlge', 'r8169', 'S2IO',
+              'sky2', 'starfire', 'tehuti', 'tg3', 'typhoon', 'via-velocity',
+              'vxge', 'gianfar', 'ehea', 'stmmac', 'vmxnet3']
+  $grep_test = join($drivers, ' -e ')
+
+  if (::osfamily == 'RedHat' and ::use_vlan_splinters) {
```
aglarendil: looks like you missed '$' sign

xarses:
Tested using libvirt and e1000 nic type. http://paste.openstack.org/show/50865/. This resolves the problem and healt tests pass.

The [ -f /dev/sys ... ] test was removed as all files there are links and ethtool will just return -1 if it's not suppored and will fail the grep test if not an expected driver. Testing shows that only the physical interfaces are enumerated as it is.

The $::osfamily test was replaced with a kernelmajversion and a regex to match kernels before 3.3

PR #807, opened 2013-11-07
increase up timeout

xdel:
Should be tested thoroughly. User already stuck with it (Centos 3.2) with physical switch.

### DIFF
--- a/deployment/puppet/l23network/lib/puppet/type/l3_if_downup.rb
+++ b/deployment/puppet/l23network/lib/puppet/type/l3_if_downup.rb
@@ -39,7 +39,7 @@ Puppet::Type.newtype(:l3_if_downup) do
     end
 
     newparam(:sleep_time) do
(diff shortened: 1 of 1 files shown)


xenolog:
-1

20sec per iface will lead to 7*20 on 4-interfaces topology.

xdel:
Ehm, generally code seems like ceiling but not just a fixed amount of time. Anyway, I`ll appreciate any proposal which will help at the same point.

xenolog:
this timeout affect someone  1st time while more than half of year. 
I think it's very custom case. And we don't put custom workarounds for all cases.
Instead this we should create possibility for enabling some workarounds for some interfaces by CLI.

xenolog:
As we investigated -- it's another timeout.

xdel:
Will be deprecated by more intelligent link state checker.

PR #816, opened 2013-11-11
syntax fix

teran:
Syntax fix to use double quotes instead of single quotes


teran:
Covered with https://github.com/Mirantis/fuel/pull/818

PR #817, opened 2013-11-11
Fix #1249999

bogdando:
```
This PR should be tested both with https://mirantis.jira.com/browse/OSCI-871.
Will fail in case of non fixed neutron packages and OCF scripts.

Closes bug: #1249999
```


xenolog:
DO NOT MERGE it. I make some addition and pull all it by one PR.

PR #818, opened 2013-11-11
debugs to puppet added according to diagnose wild ip bug

teran:
- Added debug entries to puppet network configuration libs
- syntax fix in mon.pp

**please do not merge until flashing**


teran:
Not necessary. #824 fixes wild IP, so no debug info needed in logs.

PR #820, opened 2013-11-11
fix Horizon problem on Ubuntu/simple

xenolog:
PRD-2549 for Ubintu


aglarendil:
-1

you shiould better find who is creating this file and run this command from wsgi_user

aglarendil:
+2

PR #821, opened 2013-11-11
Swift output fix

Zipfer:
BUG: https://bugs.launchpad.net/fuel/+bug/1248651
tested only with rspec


dmitryilyin:
Parsing commands output with regeps is SOOO fun, especially when output format changes all the time and there is no rest api.

aglarendil:
+2

PR #839, opened 2013-11-14
add ntp to nailgun

mattymo:
add resolv.conf and dhcp-enter-hook for nailgun
make ntp template for fuel master server
add ntp config template to openstack::clocksync
make ntp template for fuel master server


aglarendil:
+2

PR #849, opened 2013-11-14
[3.2] [PRD-2485] Add support for vlan splinters

xarses:
Add option use_vlan_splinters to site.pp. If enabled and kernelmajversion
 < 3.3 then vlan splinters will be enabled if the driver for an bridge
 interface is present in the devices list in l23network::l2::splinters.

This behavior is disabled by default as not all installs may suffer from this
 and should only be enabled when necessary.

### DIFF
--- a/deployment/puppet/l23network/manifests/l2/port.pp
+++ b/deployment/puppet/l23network/manifests/l2/port.pp
@@ -54,6 +54,12 @@ define l23network::l2::port (
       interface_properties => $interface_properties,
       skip_existing => $skip_existing
     }
+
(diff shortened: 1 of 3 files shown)


angdraug:
+1

deployment/puppet/l23network/manifests/l2/port.pp
```diff
@@ -54,6 +54,12 @@
       interface_properties => $interface_properties,
       skip_existing => $skip_existing
     }
+
+    if ($::kernelmajversion =~ /^(2.\d|3.[0-2])/ and $::use_vlan_splinters) {
```
xenolog, line 58: Please do not use global vars. pass this as parameter of puppet resource

aglarendil:
implemented by @xenolog

PR #850, opened 2013-11-14
[3.2] Fix ceph, radosgw init issues

xarses:
This patch forces $::operatingsystem == Ubuntu to use the debian init script
 instead of the upstart script for radosgw. This is necessary because the
 upstart script requires a parameter that breaks puppet's upstart provider.
 this resolves #1239348.

This patch more strickly sets the dependancy for service ceph in init.pp.
 This some errors found in #1250122.

Closes-bug #1250122
Closes-bug #1249348


aglarendil:
+2

PR #851, opened 2013-11-15
Havana mopup

aglarendil:
This one removes some fixes not needed in havana branch: e.g. debug logging for sporadical issues with IP assignment  and horizon css files recompilation which have been already fixed.


aglarendil:
+2

PR #864, opened 2013-11-21
Fix ubuntu static configuration snippet

aglarendil:
Closes-bug: 1248954


aglarendil:
+2 tested and verified

PR #867, opened 2013-11-22
anti-execution-expired patches

xenolog:
- CIBADMIN without pipes
- l3-agent
- dhcp-agent
- metadata-agent
- ovs-agent
- galera


monroid:
retest this please

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
```diff
@@ -17,7 +17,44 @@
...
+    salt = rand(0x0010000)
+    tmpfile_name = "/tmp/cib-#{Process.pid}-#{tstamp}-#{salt}.xml"
+    cmd = "#{command(:cibadmin)} -Q"
+    if Puppet[:debug]
+      statusfile_name = "/tmp/crm_status-#{Process.pid}-#{tstamp}-#{salt}.txt"
+      begin
+        #todo: more carefuly calculate or parametrize timeout
+        Timeout::timeout(30) do
+          `crm status > #{statusfile_name}`
+          rc = $?.exitstatus
```
Zipfer, line 30: Open3 for get exit status or handle it in puppet native commands

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
```diff
@@ -17,7 +17,44 @@
...
+          raise Puppet::Error("Command '#{cmd}' returns rc=#{rc}")
+        end
+      end
+    rescue Timeout::Error
+      #todo: Check XML, and no raise if it valid.
+      raise Puppet::Error("Command '#{cmd}' execution expired.")
+    end
+    if ! File.exists?(tmpfile_name)
+      raise Puppet::Error("TEMP file '#{tmpfile_name}' not found.")
+    end
```
Zipfer, line 55:
``` ruby
raise Puppet::Error("TEMP file '#{tmpfile_name}' not found.") unless File.exists?(tmpfile_name)
```

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
```diff
@@ -241,10 +278,10 @@ def start
     crm('resource', 'start', get_service_name)
     debug("Starting countdown for resource start")
     debug("Start timeout is #{@service[:start_timeout]}")
-    Timeout::timeout(@service[:start_timeout],Puppet::Error) do
+    Timeout::timeout(2*@service[:start_timeout],Puppet::Error) do
```
Zipfer: why we can't change this value in the start_value type? it's not obvious

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
```diff
@@ -17,7 +17,44 @@
...
+        end
+      rescue Timeout::Error
+        raise Puppet::Error("Command 'crm status' execution expired.")
+      end
+    end
+    begin
+      #todo: more carefuly calculate or parametrize timeout
+      Timeout::timeout(30) do
+        # execute cibadmin and store xml in tmp file
+        `#{cmd} > #{tmpfile_name}`
```
Zipfer, line 43: linux commands shouldn't be used if you want handle stderr and stdout

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
```diff
@@ -17,7 +17,44 @@
   has_feature :enableable
   has_feature :ensurable
   def self.get_cib
-    raw, _status = dump_cib
+    tstamp = Time.new.strftime("%Y%m%d-%H%M%S")
+    salt = rand(0x0010000)
+    tmpfile_name = "/tmp/cib-#{Process.pid}-#{tstamp}-#{salt}.xml"
```
Zipfer, line 22: why we should use file? 
old style was more correct you can add retries for it

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
xenolog, line 30: what you mean? we do not use open3 here.

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
xenolog, line 43: we do not handle std\* by ruby code here

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
xenolog, line 22: 1. we forget handle stdout/stderr by ruby code.
2. for debug purpose

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
Zipfer, line 30: we should use open3 or puppet 'commands' methods for get output

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
Zipfer, line 43: for example:

``` ruby
rc = $?.exitstatus
if rc != 0
```

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
Zipfer, line 22: for debug you can use debug(output)and see it in puppet log with debug output
or (but I don't understand why) save it 

``` ruby
File.open(target, "w+") do |f|
  f.write(content)
end
```

xenolog, line 22: see bellow

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
xenolog, line 30: one of main aims this patch -- forget use pipes. Open3 use pipes between chield and main process, and can't use here.

Zipfer, line 30: ok, you can use this method in your own deploy, but it should not be in the master

xenolog, line 30: it's only you personal opinion. nothing more.

mattymo, line 30: We should not be writing to temporary or unusual log files during deployment. Ivan's view on this matter is we should append all debug info to the puppet log, not to separate files. Keeping logs in one place for puppet run makes it easier to identify order of operations. 

Additionally, it's fine when troubleshooting to add logs to external files as a one-time hack, but code like this should not go in master.

dmitryilyin, line 30: Sergey's idea is to use temporary xml file. He runs command to get cib xml dump and redirects it to this file. Then just opens this file and reads it's content. He does it to solve problems we have been haveing with popen3 and debugging is only side effect of this change. Using temporyry files in synchronus process? Why not, it's ok.
Yes, we have already found out there popen3 promlems are causes by missing read call, but Sergey didn't know about it when this code was written and he used another solution.
Ivan suggests to use popen3 with read to get xml content and use Puppet.debug to log debugging information.
It's a good solution too. But xml dumps are a bit to large to be logged so Ivan have made logging of crm status in puppet debug log.

So in my opinion both solutions are ok, but it would be better to use popen3 because it's already used pretty much everywhere.

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
dmitryilyin, line 55: Absolutly same code but it Pythonic way. Why should you force you favorite style to other people? Ruby doesn't have any official code style guidelines (Python does). So any code that works is a correct one. Like in the Perl way.

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
dmitryilyin: Yes, i have no idea why does he double the timeout too. I hope he has his reasons to do so.

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
dmitryilyin, line 43: Yes, if you are using command eithin ` characters you can get exitcode by $?.exitstatus
Or you can use resurl = system(cmd) to get success (true/false)
Or you can use Puppet's cmd system to get both error processing and debug output.

All ways are correct but third is better.

Ruby is like Perl, there is more then one way to do it.

mihgen:
I hope it all works .. such a nightmare hell

deployment/puppet/corosync/lib/puppet/provider/service/pacemaker.rb
Zipfer, line 55: https://github.com/bbatsov/ruby-style-guide
the first link in google ;)

PR #870, opened 2013-11-22
add read function to IO

Zipfer:
-Fix Illegal seek bug
  -https://bugs.launchpad.net/fuel/+bug/1250776

Please remove cleanup from all manifests and dependency


aglarendil:
+2

PR #874, opened 2013-11-25
Add static udev everywhere

xdel:
https://review.openstack.org/#/c/58260/

please test


mattymo:
Normally so many seds and trs are really a waste of pipes and a tangle of code, but the results are consistent and I can't find any obvious flaws here. I was able to rewrite in a few minutes with a while read loop that reads from a pipe of awk, but it still looked even worse than what Andrey wrote. I think with about an hour of work I could do it a little better, but it's only going to improve readability here. The only recommendation I can make is to add a comment indicating what happens here, like "#Systematically identifying interfaces so cobbler does not stomp them"

mihgen:
can't we use simplest python script instead of complicated bash then?

mihgen:
+1
as it is tested and looks like it works.

PR #877, opened 2013-11-25
	Flush IPs on all interfaces that are members of a bond

alexeyklimenok:
```
- This is needed to prevent the situation when Puppet
is trying to assign an IP to bond, which is already in
use on some of bonded interfaces.
```


xenolog:
good idea.
but this patch affects only realtime situation. 

what happens after reboot?

xenolog:
-2

https://github.com/xenolog/l23network
Bonding Section.

xarses:
Why did we merge something with -2 review?

aglarendil:
A, Andrew, this was my fault. I accidently pushed 'merge' instead of 'close' :-) It is already reverted.

PR #878, opened 2013-11-26
No splash for Ubuntu

xdel:
https://bugs.launchpad.net/bugs/1255026


xdel:
abandoning, further work in the https://bugs.launchpad.net/fuel/+bug/1259245

PR #884, opened 2013-11-29
Nicira NSX integration POC

dukov:
Design doc 
https://docs.google.com/a/mirantis.com/document/d/1QW6VNWlO7RHVL9R_kv4xFfnyjc2FTfXPRbSta1Sh8jc/edit


deployment/puppet/neutron/manifests/plugins/nicira.pp
```diff
@@ -0,0 +1,98 @@
...
+    } ->
+    file {'/etc/neutron/plugins/openvswitch':
+      ensure  => directory,
+      mode    => '0755',
+    } ->
+    file { '/etc/neutron/plugin.ini':
+      ensure  => link,
+      target  => '/etc/neutron/plugins/nicira/nvp.ini',
+    } ->
+    neutron_plugin_nicira {
```
bogdando, line 76: Quantum/Neutron agents inherit all logging options from server, but what about NVP logging?

dukov, line 76: Logging in NVP is implementer as in OVS plug-in

bogdando, line 76: I believe we should use it to include the NVP into Fuel common logging scheme

PR #888, opened 2013-11-29
Fix start/cleanup race condition

xenolog:
Do service cleanup on the particluar node
if it is not installed on the node.


aglarendil:
+2

aglarendil:
tested on CentOS

PR #889, opened 2013-11-29
Fix start/cleanup race condition

aglarendil:
Do service cleanup on the particluar node
if it is not installed on the node.


aglarendil:
Forwardport from 3.2-fixes

PR #894, opened 2013-12-03
Fuel utils and quantum keys

dmitryilyin:
Experimantal FDP table cleaning tool.

Don't merge!

Installs ssh keys to allow access between all computes and controllers which may not be what you want security-wise.

OCF scripts also should be made templates to avoid key path hardcoding.


xdel:
Please prepare with 's/quantum/neutron/g' and propose for review. Works for me on current Havana master.

PR #895, opened 2013-12-03
Passed zero size disks in pmanager

kozhukalov:


yoojinl:
+

aglarendil:
pull request system test is not relevant for this change. Could you please confirm that ISO built with this code works fine?

adidenko:
I've tested this change on a live fuel-3.2.1-85 without building iso. It worked OK, I've deployed CentOS env just fine.

aglarendil:
+1

PR #898, opened 2013-12-03
OVS dp is back

xdel:
Depends on: 
https://review.openstack.org/59692
https://review.openstack.org/59693

Needs to be cherry-picked to the master.


xdel:
postponed for 4.0

PR #900, opened 2013-12-03
Fix wget timeout in RHN satellite key download

mattymo:
wget http://$redhat_management_server/pub/RHN-ORG-TRUSTED-SSL-CERT -O $mycert takes 20 minutes (60s TCP timeout \* 20 retries) if there is DNS resolution but no internet access.

Rewrote Cobbler snippet to set a reasonable timeout if the connection can't be made.
https://bugs.launchpad.net/fuel/+bug/1257285


aglarendil:
+2

PR #901, opened 2013-12-03
return of the out-of-tree package

xdel:
Depends on https://review.openstack.org/#/c/59693/ and OSCI-946


mihgen:
Dependent pull request https://review.openstack.org/#/c/60150/ was merged. Should we merge this one right away?

PR #904, opened 2013-12-03
Quantum services management

xdel:


mihgen:
Started verification by manual replacement of files affected and re-run of puppet on already installed env (which was initially affected).

mihgen:
Run was successful. Killed quantum services, rm -rf /etc/quantum, uninstalled l3agent and quantum-server, trying to run puppet again..

mihgen:
After such a hacky check Quantum API is not available:
err: Could not prefetch quantum_net provider 'quantum': Can't prefetch net-list. Quantum or Keystone API not availaible.

mihgen:
Fresh installation passed controller deployment, and l3agent started:
root@node-4:~# ps aux|grep l3
quantum  15356  0.4  1.9 103752 29920 ?        Ss   06:33   0:00 python /usr/bin/quantum-l3-agent --config-file=/etc/quantum/quantum.conf --config-file=/etc/quantum/l3_agent.ini --log-file=/var/log/quantum/l3-agent.log
root     18333  0.0  0.0   9384   928 pts/3    S+   06:36   0:00 grep --color=auto l3
root@node-4:~# ip netns list
qrouter-81c21b95-6e5c-4f4b-a76d-31e889e8e170
qdhcp-a1573797-7fc8-4ad9-9640-e1b493c0be64

No errors in l3agent logs found. Considering this patch as good to go.

deployment/puppet/quantum/manifests/agents/dhcp.pp
```diff
@@ -20,6 +20,19 @@
          mode    => 644,
          before  => Package['quantum-dhcp-agent'],
        }
+    } else {
+       file { '/etc/init/quantum-dhcp-agent.override':
+         replace => 'no',
+         ensure  => 'present',
+         content => 'manual',
+         mode    => 644,
+         before  => Package['quantum-dhcp-agent'],
```
xenolog, line 29: I think we have not to use "before", "require" in our manifests.
We should use '->' relationship between resources.
It preserve us from old puppet bug

PR #905, opened 2013-12-04
arrow style for conditions

xdel:


mihgen:
deployed 3 environments with one controller in each. Verified that namespaces created and l3 agent running.

xdel:
abandoning

PR #906, opened 2013-12-04
Add neutron support and ciurros for murano

dmitryilyin:
Update for murano from Max Mazur
Don't merge until tested!


aglarendil:
+2

PR #909, opened 2013-12-04
FIX: bug/1257904

xenolog:
puppet send non-boolean value instead true/false


deployment/puppet/l23network/lib/puppet/provider/l2_ovs_port/ovs.rb
```diff
@@ -55,7 +55,8 @@ def create
       end
     end
     # enable vlan_splinters if need
-    if @resource[:vlan_splinters]
+    if @resource[:vlan_splinters].to_s() == 'true'  # fucking puppet send non-boolean value instead true/false
```
xarses: Please remove the f-word from the comment

ashaposhnikov:
+1 if you'll remove f-word

xarses:
Confirmed that it removes the problem, update the comment.

aglarendil:
+2

PR #910, opened 2013-12-05
Add fuel-utils to fuel

xenolog:


mattymo:
Sergey, why rewrite daemonize class? It already exists and we have packages available for it. It's available under MIT license which is very free and we can rewrite and redistribute with no issues, but we shouldn't need to.

PR #911, opened 2013-12-05
4.0 ex expired non quantum

dmitryilyin:


deployment/puppet/cobbler/templates/scripts/pmanager.py
```diff
@@ -304,8 +304,12 @@ def pvs(self):
                              begin_size, end_size, self.unit))
                 self.kick("partition {0} "
                           "--onpart=$(readlink -f /dev/{2}){3}{4}"
+<<<<<<< HEAD
```
bogdando: yay! bad merging

xarses:
looks like a rebase is needed

aglarendil:
+2

xarses:
Umm, why on earth did we merge this with the broken commit chain?

PR #916, opened 2013-12-09
package order for a simple mode

xdel:


xenolog:
I already port this changes in case Bug#1257806

aglarendil:
Anyway, this one contains better ordering. Lets' merge different bugs in separate requests.

PR #922, opened 2013-12-09
Updated cirros.pp

teselkin:
This update allows Murano Demo VM to be installed and registered in Glance.


aglarendil:
we should not use this as image uploading is now made by orchestrator

PR #924, opened 2013-12-09
avoid kmod-openvswitch installation for non-2.6 kver

xdel:
Proposed-for: https://blueprints.launchpad.net/fuel/+spec/cisco-ucs-imp
Milestone: 4.1
Deprecates: VLAN splinters usage
Depends-on: bound removal between kmod-openvswitch and openvswitch packages, nailgun kernel selection passthrough

### DIFF
--- a/deployment/puppet/l23network/manifests/l2.pp
+++ b/deployment/puppet/l23network/manifests/l2.pp
@@ -16,7 +16,7 @@ class l23network::l2 (
        before => Package[$::l23network::params::ovs_packages],
      }
     }
-    if $::operatingsystem == 'Centos' {
+    if ( $::operatingsystem == 'Centos' and $::kernelmajversion == '2.6' ) {
      package { 'kmod-openvswitch':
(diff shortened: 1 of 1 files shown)


xarses:
How does this depercrate vlan splinters?

xdel:
This exact commit does not deprecate them but usage of 3.10+ kernel does - it will be replacement for current splinters-inavoidable cases. We will have no overhead for any imaginable vlan ranges and slightly broader HCL paying by kABI breakage. Old 2.6.32 will remain as a default selection of course.

xarses:
At least one user has indicated that the 3.10 kernel causes panic's on their hardware (Slightly Older HP Blade) Moving to fedora kernel Isn't a good option for everyone. This also indicates that further hardware testing must be done prior to accepting the new kernel

xdel:
It should not be default selection but only for models having unbeatable issue with connectivity in VLAN isolation mode.

PR #932, opened 2013-12-11
Adds ceilometer-alarm support

r7vme:
- simple and HA mode
- some refactoring in the alarm manifests

### DIFF
--- a/deployment/puppet/ceilometer/files/ocf/ceilometer-alarm-evaluator
+++ b/deployment/puppet/ceilometer/files/ocf/ceilometer-alarm-evaluator
@@ -1,0 +1,321 @@
+#!/bin/sh
+#
+#
+# OpenStack Ceilometer Alarm Evaluator Service (ceilometer-alarm-evaluator)
+#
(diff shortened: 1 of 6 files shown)


deployment/puppet/ceilometer/files/ocf/ceilometer-alarm-evaluator
```diff
@@ -0,0 +1,321 @@
...
+# Initialization:
+
+: ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
+. ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
+
+#######################################################################
+
+# Fill in some defaults if no values are specified
+
+OCF_RESKEY_binary_default="ceilometer-alarm-evaluator"
```
xdel, line 34: full path would be better

r7vme, line 34: hmm, is it necessary? I do not use full path, because

1) anyway it checked by "check_binary $OCF_RESKEY_binary" function, that split full path and tried to check with "which" https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/ocf-binaries.in#L72
2) for different distributions binary could be installed in different path

